PR #2, opened 2026-07-26
ci: add OpenCodeReview workflow

yingdi-shan:
## Summary

- add automated OpenCodeReview reviews for non-draft pull requests
- use `ubuntu-latest` and track `alibaba/open-code-review@main`, including its default latest CLI
- cancel stale reviews when a pull request is updated
- grant only read access to contents and write access to pull-request reviews

## Configuration

Define these repository Actions secrets after merge:

- `OCR_LLM_URL`: LLM API endpoint
- `OCR_LLM_AUTH_TOKEN`: LLM API key
- `OCR_LLM_MODEL`: model name

The optional `OCR_LLM_USE_ANTHROPIC` repository variable selects the Anthropic protocol; it defaults to `false` for OpenAI-compatible endpoints.

## Security

The workflow uses `pull_request_target` for fork pull requests. Tracking `alibaba/open-code-review@main` intentionally follows upstream changes without pinning.

### DIFF
--- a/.github/workflows/open-code-review.yml
+++ b/.github/workflows/open-code-review.yml
@@ -1,0 +1,34 @@
+# Required Actions secrets:
+#   OCR_LLM_URL: LLM API endpoint
+#   OCR_LLM_AUTH_TOKEN: LLM API key
+#   OCR_LLM_MODEL: model name
+
+name: OpenCodeReview
+
+on:
+  # Required for fork PRs to access the base repository's OCR secrets. The OCR
+  # action fetches the PR diff without executing contributor-controlled code.
+  pull_request_target:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+concurrency:
+  group: open-code-review-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  code-review:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Review pull request
+        uses: alibaba/open-code-review@main
+        with:
+          llm_url: ${{ secrets.OCR_LLM_URL }}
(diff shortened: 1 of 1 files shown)


LSX-s-Software (commented):
Why not use [Open Code Review's official workflow example](https://github.com/alibaba/open-code-review/blob/main/examples/github_actions/ocr-review.yml)?

yingdi-shan:
> Why not use [Open Code Review's official workflow example](https://github.com/alibaba/open-code-review/blob/main/examples/github_actions/ocr-review.yml)?

Good point. This workflow follows the official example’s core approach. It only omits the example’s /ocr review comment trigger since Open Code Review ignores prior context. https://github.com/alibaba/open-code-review/issues/59

guozy18 (commented):
Could we pin the action to a full commit SHA and lock `ocr_version` to an exact version? It's two extra lines of config and removes two mutable execution sources. GitHub calls out the full-length SHA as the only immutable way to reference an action ([secure use reference](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)).

>
> ```yaml
> uses: alibaba/open-code-review@<audited-full-sha>
> with:
>   ocr_version: "<exact-version>"
> ```

yingdi-shan:
> Could we pin the action to a full commit SHA and lock `ocr_version` to an exact version? It's two extra lines of config and removes two mutable execution sources. GitHub calls out the full-length SHA as the only immutable way to reference an action ([secure use reference](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)).
> 
> >

You’re right that pinning provides stronger supply-chain security, which matters because this pull_request_target workflow receives secrets. The trade-off is maintenance: fixes, model compatibility updates, and new OCR capabilities would require explicitly updating both pins. The current configuration deliberately prioritizes tracking the latest Open Code Review release.

LSX-s-Software (approved):

PR #4, opened 2026-07-26
Track latest E2B clients in compatibility tests

LSX-s-Software:
## Summary

- install the latest stable E2B CLI, Python SDK, and Code Interpreter SDK in CI
- resolve the latest TypeScript SDK at test time instead of using a pinned dependency and lockfile
- use `e2b_000000` as the local placeholder API key across E2E suites and documentation

## Motivation

The compatibility suites previously tested fixed E2B client versions, which meant upstream compatibility regressions were only detected after manually updating those pins.

This change intentionally makes the suites follow the latest stable client releases. The TypeScript SDK is installed with `e2b@latest` while ignoring lockfiles, and the Python packages are upgraded on each CI run.

The placeholder API key now also follows the E2B key format expected by current clients while remaining a non-secret value for local AgentENV deployments.


yingdi-shan (approved):
LGTM.

PR #14, opened 2026-07-27
Expose host interaction IP in custom extension API

LSX-s-Software:
This pull request adds support for passing the per-runtime IPv4 address (`hostInteractionIp`) to the custom extension hooks for sandboxes. This enables extensions to receive the current interaction IP for each sandbox instance, which can change after pause/resume operations.


yingdi-shan (approved):
LGTM.

PR #19, opened 2026-07-27
feat: expose snapshot rootfs image references

### LINKED ISSUE #16: Prevent snapshot deletion from removing shared OCI manifests
## Summary

Snapshot image publications can produce identical OCI manifest digests when two publications reference the same OverlayBD layer set.

AgentENV currently records the manifest digest for each publication and deletes the manifest by digest when rolling back or deleting a snapshot. In OCI registries, multiple tags may point to the same content-addressed manifest. Deleting that manifest by digest can therefore invalidate every tag that references it, including tags owned by other snapshots.

## Affected code

- `src/snapshot/repository/backends/common/acr/manifest.rs`
- `src/snapshot/repository/backends/common/acr/publisher.rs`
- `src/snapshot/repository/backends/oss/repository.rs`

## Current behavior

The generated OCI manifest contains:

- a deterministic minimal OCI config descriptor;
- the ordered OverlayBD layer descriptors.

It does not contain the snapshot ID, publication tag, or disk subject.

Therefore, two publications with the same layer descriptors generate identical manifest bytes and the same manifest digest.

For example:

```text
agentenv-snapshot-S1 ─┐
                      ├── sha256:M
agentenv-snapshot-S2 ─┘
```

When snapshot `S1` is deleted, AgentENV executes:

```text
DELETE /v2/<repository>/manifests/sha256:M
```

Because `sha256:M` is shared, the registry may remove the manifest referenced by both tags. `agentenv-snapshot-S2` then becomes unavailable even though snapshot `S2` still exists.

This is especially easy to trigger when publishing snapshots whose rootfs consists entirely of remote layers, but it can occur for any two publications that produce the same config and layer descriptors.

## Expected behavior

Each snapshot rootfs or attached-drive publication should own a distinct manifest digest, even when its config and OverlayBD layer descriptors are identical to another publication.

Deleting or rolling back one snapshot publication must not invalidate another snapshot's image tag.

Layer and config blobs should remain content-addressed and shared; only the small manifest object needs to be publication-specific.

## Proposed fix

Add the exact publication tag to the top-level OCI manifest annotations:

```json
{
  "annotations": {
    "io.agentenv.snapshot.tag": "agentenv-snapshot-<snapshot-id>"
  }
}
```

Attached-drive manifests should use their complete generated tag:

```text
agentenv-snapshot-<snapshot-id>-drive-<drive-id>-<hash>
```

Including the publication tag makes the manifest bytes and digest unique per publication while preserving deduplication of:

- existing remote OverlayBD layers;
- newly uploaded delta layers;
- the minimal OCI config blob.

No layer data needs to be duplicated or re-uploaded.

## Cleanup reliability

Snapshot deletion should remove external registry publications before deleting the durable snapshot catalog record.

If registry cleanup fails, the catalog record and its persisted publication metadata must remain available so deletion can be retried. Invalid persisted publication metadata must return an error rather than being treated as successful cleanup.

Manifest deletion should remain idempotent: a registry `404 Not Found` is considered successful so partially completed multi-publication cleanup can be retried safely.

huajq:
## Summary

Expose the existing source-registry snapshot rootfs publication as an optional `imageRef` in snapshot API and CLI responses.

Image publication remains part of snapshot creation and still requires the OSS snapshot backend with `[snapshot.image_publish].enabled = true`. This PR does not add a separate export endpoint or CLI command.

Fixes #16.

## Changes

- Add an optional `imageRef` to snapshot create, GET, and list responses.
- Print the published image after `aenv snapshot create`.
- Add an `IMAGE REF` column and optional JSON field to `aenv snapshot list`.
- Select rootfs publications by their exact deterministic tag so attached-drive publications cannot be exposed accidentally.
- Publish an OCI manifest and snapshot tag even when the rootfs contains only existing remote layers.
- Keep remote OverlayBD layers referenced by digest without downloading or uploading them again.
- Add the exact publication tag to OCI manifest annotations so otherwise identical snapshot publications receive distinct manifest digests.
- Delete registry publications before discarding their persisted snapshot metadata, allowing failed cleanup to be retried.
- Leave stale alias cleanup to the existing lookup/bind path to avoid races with concurrent alias rebinding.
- Reject invalid persisted registry publication metadata instead of silently discarding cleanup information.
- Reject HTTPS-to-HTTP registry redirects and limit redirect chains to 10 hops.
- Update snapshot image publication configuration, concepts, and CLI documentation.

## Compatibility

- `imageRef` is optional and omitted when no rootfs image was published.
- Existing snapshot records remain readable.
- Older clients can ignore the new response field.
- Standard OCI images converted to local OverlayBD layers retain the existing object-storage fallback.
- No new endpoint, background task, dependency, or persistent schema migration is introduced.

## Performance

- Existing remote layers are referenced by digest and are not re-uploaded.
- Local delta blobs retain digest-based existence checks and upload deduplication.
- All-remote publication performs only registry metadata requests and a manifest PUT.
- Snapshot GET/list performs only an in-memory publication lookup and does not access the registry.
- Layer and config blobs remain content-addressed and shared; only manifests are publication-specific.

## Testing

Validated on the remote Linux test host:

```text
cargo test -p agentenv --lib
640 passed, 0 failed, 2 ignored

cargo test -p aenv
23 passed, 0 failed
```

Additional checks:

```text
cargo clippy -p agentenv -p aenv --all-targets -- -D warnings
cargo fmt --all -- --check
git diff --check
```

The tests cover:

- all-remote publication through the existing fake registry;
- remote-only publication without layer uploads;
- publication-specific manifest digests;
- rootfs versus attached-drive publication selection;
- optional API and CLI `imageRef` serialization;
- invalid persisted publication metadata;
- HTTPS redirect enforcement;
- existing blob deduplication, chunked upload, and manifest deletion behavior.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -129,12 +129,17 @@ p2p_enabled = true
 #   - "oss": store committed snapshots in object storage.
 #     Requires [backend.oss] to be configured.
 #
-# Advanced: enable only when source images already live in an overlaybd-capable
-# registry and AENV has push credentials. When enabled, AENV publishes
-# each new snapshot delta as a new overlaybd image tag in the original source
-# registry, so the snapshot can also be addressed and reused as an image version
-# there. Unsupported source images fall back to the configured snapshot
-# repository.
+# Advanced: only takes effect with repository_backend = "oss". Enable only when
+# source images already live in an overlaybd-capable registry and AENV has push
+# credentials (from ~/.docker/config.json). When enabled, AENV publishes each
+# snapshot rootfs as a new overlaybd image tag "agentenv-snapshot-{snapshot_id}"
+# in the original source registry: existing remote layers are referenced by
+# digest and only new delta layers are uploaded, so the snapshot can also be
+# addressed and reused as an image version there. The published reference is
+# returned as `imageRef` in snapshot APIs. Memory and VM-state artifacts always
+# stay in the snapshot repository. Locally converted (non-overlaybd) source
+# images fall back to object storage; incompatible remote-backed layouts fail
+# the publish instead of mixing sources.
 # [snapshot.image_publish]
 # enabled = true
 
--- a/config/oss_default.toml
+++ b/config/oss_default.toml
@@ -87,12 +87,18 @@ auto_resume_min_sandbox_timeout_secs = 300
 repository_backend = "oss"
 p2p_enabled = true
 
-# Advanced: enable only when source images already live in an overlaybd-capable
-# registry and AgentENV has push credentials. When enabled, AgentENV publishes
-# each new snapshot delta as a new overlaybd image tag in the original source
-# registry, so the snapshot can also be addressed and reused as an image version
-# there. Unsupported source images fall back to the configured snapshot
-# repository.
+# Advanced: only takes effect with repository_backend = "oss". Enable only when
+# source images already live in an overlaybd-capable registry and AgentENV has
+# push credentials (from ~/.docker/config.json). When enabled, AgentENV
+# publishes each snapshot rootfs as a new overlaybd image tag
+# "agentenv-snapshot-{snapshot_id}" in the original source registry: existing
+# remote layers are referenced by digest and only new delta layers are
+# uploaded, so the snapshot can also be addressed and reused as an image
+# version there. The published reference is returned as `imageRef` in snapshot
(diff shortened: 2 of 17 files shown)


yingdi-shan (approved):
LGTM.

PR #23, opened 2026-07-27
feat: support virtual-host addressing for S3-compatible snapshot back

davidmyriel:
## What problem does this solve?

I tried to point the OSS snapshot backend at an S3-compatible provider that requires virtual-host bucket addressing and found there was no way to make it work.

The backend picks the addressing style automatically in `detect_addressing_style`: Alibaba OSS endpoints and endpoints that already contain the bucket name get virtual-host style (`https://bucket.example.com/key`), and everything else falls back to path style (`https://example.com/bucket/key`). That fallback is the problem — a growing set of providers only reliably support virtual-host addressing (AWS S3 buckets created after the path-style deprecation, Cloudflare R2, Tigris), and for those the auto-detection guesses wrong with no way to override it. The result is confusing request failures at runtime rather than anything that points at addressing style.

## What this PR does

Adds an optional `addressing_style` field to `[backend.oss]`:

```toml
[backend.oss]
endpoint = "https://t3.storage.dev"
bucket = "agentenv-snapshots"
region = "auto"
addressing_style = "virtual"   # or "path"
```

The behavior is strictly opt-in:

- **When set**, the explicit value wins and is passed straight through to the object-store client.
- **When unset**, the existing `detect_addressing_style` auto-detection runs exactly as before — zero behavior change for current Alibaba OSS / MinIO setups.

The plumbing is small: a new `OssAddressingStyle` config enum, a mapping in `NormalizedOssConfig`, and an override-or-detect resolution in `OssClient::new`.

## Testing

- Unit tests covering: the detection fallback for generic endpoints, the existing Aliyun/bucket-in-endpoint detection, the override taking precedence, and the config mapping.
- Verified end-to-end against a real Tigris bucket with the included `tigris_smoke` example, which drives the same `build_object_store_operator` path the backend uses: write → read-back → verify → delete round-trip succeeds with `addressing_style = "virtual"`.
- Also confirmed at the wire level (via a local Host-header echo endpoint) that the two styles genuinely produce different requests — `Host: bucket.endpoint` vs `/bucket/key` on the path — so the override reaches the wire rather than being silently dropped.
- `cargo fmt`, `cargo clippy -- -D warnings`, and `cargo test -p object-store-operator` are clean.

## Docs

- `docs/src/configuration/reference.md`: new row in the `[backend.oss]` table plus a short worked example.
- `docs/src/getting-started/on-demand-loading.md`: one-line pointer after the OSS example config, since anyone following that guide with a virtual-host-only provider would otherwise hit an opaque failure.
- Commented-out example in `config/default.toml`.

Happy to drop the smoke example if you'd rather keep `examples/` lean


src/snapshot/repository/backends/oss/mod.rs
```diff
@@ -64,6 +64,7 @@ impl OssBackend {
             config.region().to_string(),
             config.prefix().to_string(),
             config.credential_source(),
+            config.addressing_style(),
```
guozy18, line 67: I think we also need to pass this setting through to OverlayBD. Snapshot metadata uses OssClient, but remote managed layers are read through storage/overlaybd/src/backend/oss.rs, which still auto-detects the addressing style. On a virtual-host-only provider, uploads may work while restoring a sandbox still fails. Could we propagate the same override into the generated OverlayBD config and add a test for remote layer reads?

src/cfg.rs
```diff
@@ -289,6 +303,8 @@ pub struct OssBackendConfig {
     pub access_key_secret: Option<String>,
     pub security_token: Option<String>,
     pub region: Option<String>,
+    #[serde(alias = "addressingStyle", alias = "addressing-style")]
+    pub addressing_style: Option<OssAddressingStyle>,
```
guozy18, line 316: This new field also needs to be added to the OssBackendConfig literal in tests/snapshot_oss_e2e_test.rs::test_oss_config. Otherwise, all-target builds and clippy will fail because the field is missing.

src/snapshot/repository/backends/oss/client.rs
```diff
@@ -34,10 +34,17 @@ impl OssClient {
...
         prefix: String,
         credential_source: CredentialSource,
+        addressing_override: Option<AddressingStyle>,
     ) -> Result<Self> {
+        // An explicit config override wins; otherwise fall back to
+        // endpoint-based detection (Alibaba OSS / bucket-in-endpoint hosts).
+        let addressing_style = match addressing_override {
+            Some(style) => style,
+            None => detect_addressing_style(&endpoint, &bucket)?,
+        };
```
guozy18: This skips detect_addressing_style when an override is set, but that function also validates the endpoint URL and host. Could we always run the validation first, then let the explicit value override the detected style?

crates/object-store-operator/examples/tigris_smoke.rs
```diff
@@ -0,0 +1,77 @@
...
+    };
+
+    println!("endpoint={endpoint} bucket={bucket} addressing_style={style:?}");
+    let op = build_object_store_operator(&config, Some(&cred))?;
+
+    let key = "agentenv-smoke/roundtrip.txt";
+    let payload = b"agentenv addressing-style smoke test".to_vec();
+
+    println!("-> write  {key}");
+    op.write(key, payload.clone()).await?;
```
guozy18: Could we use a unique key for each run? This fixed key may overwrite and then delete an existing object, and concurrent smoke tests could interfere with each other.

crates/object-store-operator/examples/tigris_smoke.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    println!("-> read   {key}");
+    let got = op.read(key).await?.to_vec();
+    anyhow::ensure!(got == payload, "read-back mismatch: object content differs");
+
+    println!("-> delete {key}");
+    op.delete(key).await?;
```
guozy18: If the read or content check fails after the write, we return before deleting the object. Could we make sure cleanup is attempted on all normal error paths?

guozy18 (changes requested):
Thanks for adding this. Being able to override the addressing style is useful for S3-compatible providers that cannot use the current default.

I'm requesting changes because the setting currently only reaches the repository client, not the OverlayBD path used to read remote snapshot layers. That means uploads can succeed while restoring a sandbox still fails. There is also a missing field in an existing integration test, plus a couple of cleanup issues in the smoke example.

One more thing: could you also update the commit message to follow the Conventional Commits format? The current message, "add tigris data", does not describe the change clearly. Something like `feat(oss): support configurable S3 addressing style` would be more appropriate. See https://www.conventionalcommits.org/en/v1.0.0/.

davidmyriel:
@guozy18 all the comments are addressed and verified via test. Should be good to go.

I dropped the smoke test from the PR as it was attracting Github bots comments.

Congrats on the project success. Would like to do some promotions and comarketing + more integration work.

storage/overlaybd/src/backend/oss.rs
```diff
@@ -31,6 +31,7 @@ struct OssBackendInner {
     credentials: CachedCredentialSource,
     default_region: String,
     default_endpoint: String,
+    addressing_override: Option<AddressingStyle>,
     timeout: Duration,
     retry_count: u32,
     cached_operators: RwLock<HashMap<OperatorCacheKey, OperatorWithCredential>>,
```
guozy18, line 37: This PR changes `OssClient::new` to take six arguments. However, the snapshot image export service recently added on main still calls it with five arguments in `image_export/service.rs`.

   After rebasing, please pass the configured override there as well:

   ```rust
   config.addressing_style(),
   ```

guozy18 (changes requested):
Thanks for the kind words, and thanks for addressing the earlier feedback. The main implementation direction now looks right, while the existing auto-detection behavior remains the default.

I still have a few requests before this is ready to merge:

1. **Rebase onto the latest main and fix the resulting build failure**

2. **Address the two remaining automated review comments**

   Both look valid to me:

   - Document both supported overrides, `"virtual"` and `"path"`, and avoid overstating which providers strictly require either style.
   - Change the remote-read test from `None` to `Some(4)` so it exercises the range GET path rather than stopping after a metadata HEAD request. It would also be useful to assert that the captured request method is GET.

3. **Do a simplification pass over the implementation and tests**

   The PR currently adds around 375 lines, with a substantial amount of overlapping test coverage. I would keep the tests that protect distinct contracts: configuration parsing, repository-client override behavior, propagation into the generated OverlayBD config, endpoint validation, and one wire-level remote-read test.

   Tests that only cover unchanged auto-detection behavior, or repeat the same override behavior at adjacent layers, can likely be consolidated. For example, the newly added repository-client auto-detection tests overlap with existing OverlayBD tests, and some of the parsing, normalization, and wire-level assertions cover the same branch more than once.

guozy18:
Thanks @davidmyriel for the original proposal and implementation.

I prepared a replacement PR based on the same requirement and updated it against the current `main` branch:

- New PR: https://github.com/kvcache-ai/AgentENV/pull/176

The new PR preserves the original default behavior, propagates the addressing style through both the snapshot OSS client and the OverlayBD runtime path, and adds focused regression coverage.

The new PR credits the original work in its description. If it is accepted, #23 can be closed as superseded.

yingdi-shan:
Resolved in PR 176#.

PR #25, opened 2026-07-28
fix(ublk): stop blocking tokio workers while waiting for ublk device nodes

hemildesai:
## Problem

`wait_for_ublk_dev` sleeps with `std::thread::sleep` and is called from the ublk daemon's async device-creation paths **without `.await`**, parking a tokio worker for the whole wait:

```rust
// storage/ublk-daemon/src/server.rs (3 call sites)
if let Err(err) = wait_for_ublk_dev(dev_id).context("wait for ublk device") { ... }
```

The daemon runs a hardcoded 4-worker runtime (`worker_threads(4)`), so concurrent cold device creations are capped at 4, each holding its worker for at least the 10 ms first retry interval.

## Note

This PR was opened in error against a repository we do not own, and has been closed. Retained only for the discussion thread.

Per review feedback, the correct fix is to remove the wait entirely rather than make it async: after `add_dev`/`start_dev` return, `ublk_ctrl_start_dev -> device_add_disk -> device_add` has already created the device node, so there is nothing to wait for.

### DIFF
--- a/storage/ublk-daemon/src/main.rs
+++ b/storage/ublk-daemon/src/main.rs
@@ -262,8 +262,17 @@ fn main() -> Result<()> {
         tracing::warn!(?err, target, "failed to raise RLIMIT_NOFILE");
     }
 
+    // Device creation does short blocking work (open/ioctl on /dev) between
+    // awaits, so a small worker pool throttles concurrent sandbox creates.
+    // Default generously and allow tuning without a rebuild.
+    let worker_threads = std::env::var("AENV_UBLK_DAEMON_WORKER_THREADS")
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+        .filter(|value| *value > 0)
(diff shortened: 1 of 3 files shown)


huang-jl:
Thanks for your observation — you're right that this synchronous sleep is problematic and shouldn't stay as-is.

That said, I think the right fix here is to remove the wait entirely, rather than converting it from a sync sleep to an async one.

The reason is that after the `add_dev` and `start_dev` ublk commands are sent to the kernel, the ublk device is already present under /dev. The kernel invokes `ublk_ctrl_start_dev -> device_add_disk -> device_add`, which guarantees that the ublkb device appears in /dev before these commands return. So there's nothing to wait for, the wait isn't just poorly implemented, it's unnecessary altogether.

hemildesai:
@huang-jl Thanks for your reply. I closed this PR for now, looking forward to exploring the project further.

Issue #8, opened 2026-07-27
Reduce Rust compile time and target artifact size


LSX-s-Software:
## Background

Several common Rust build targets in AgentENV have become expensive to compile and produce very large artifacts. This affects local development feedback loops, CI runtime, disk usage, and the cost of switching between dev/test/release profiles.

A clean full rebuild is a useful way to expose the baseline cost, but the issue is broader than any particular local cleanup workflow. Even incremental builds can be slow when touching code inside the large root `agentenv` crate, and test builds produce multiple large binaries that quickly inflate `target/`.

## Local measurements

Environment: 16 cores, 31 GiB RAM, `rustc 1.97.0`, current workspace.

The test targets below were measured with `--no-run` to isolate compile/link time from KVM, root permissions, network namespaces, or test runtime behavior.

| Target | Command | Time |
| --- | --- | ---: |
| dev server | `cargo build -p agentenv --bin server --timings` | 3m36s |
| unit tests | `cargo test -p agentenv --lib --no-run --timings` | 3m16s |
| integration tests | `cargo test -p agentenv --no-run --test integration --test orchestrator_integration --test snapshot_oss_e2e_test --timings` | 1m01s |
| release server | `cargo build --release -p agentenv --bin server --timings` | 4m49s |
| no-op rebuild | same commands immediately repeated | ~0.5s–0.8s |
| touch one root crate file | `touch src/local_store.rs && cargo build -p agentenv --bin server` | 8.4s |

`target/` size after building the common dev/test/release targets:

```text
target/                    21G
target/debug/              19G
target/debug/deps/         13G
target/debug/build/       1.1G
target/debug/incremental/ 5.0G
target/release/           2.1G
target/release/deps/      1.6G
target/release/build/     492M
```

Large artifacts observed:

```text
852M target/debug/deps/agentenv-...             # unit test binary
813M target/debug/server                        # debug server
621M target/debug/deps/orchestrator_test-...
608M target/debug/deps/snapshot_test-...
606M target/debug/deps/snapshot_attached_drive_test-...
603M target/debug/deps/fc_test-...
602M target/debug/deps/ublk_test-...
598M target/debug/deps/process_test-...
598M target/debug/deps/envd_supervisor_test-...
463M target/debug/deps/libagentenv-....rlib
202M target/debug/deps/libaws_sdk_s3-....rlib
186M target/debug/deps/libiroh-....so
116M target/debug/deps/liboverlaybd-....rlib
[...]

PR #39, opened 2026-07-28
perf: isolate benchmark and e2e dependencies

NickNYU:
## Summary

- move the four Criterion benchmarks into a dedicated `agentenv-benchmarks` workspace package
- move the Docker/MinIO OSS snapshot test into `agentenv-e2e-tests`
- keep the existing Makefile entry points while removing Criterion and test-support from the root package dev-dependencies
- update internal documentation for the relocated OSS test

Part of #8 (P1).

## Why

Cargo activates a package's dev-dependencies for library test builds even when the test target does not use them. Keeping Criterion and `agentenv-test-support` on the root package caused `cargo test -p agentenv --lib --no-run` to compile benchmark tooling, AWS SDK crates, and testcontainers.

Dedicated workspace packages preserve the benchmark/e2e targets while moving those costs behind explicit package selection.

## Measured result

Same isolated VM for both runs: Ubuntu 22.04 ARM64, 8 vCPU, 16 GiB RAM, rustc 1.97.1. Commands used a fresh target directory and a warm Cargo source cache.

| `cargo test -p agentenv --lib --no-run --locked` | Before | After | Change |
| --- | ---: | ---: | ---: |
| wall time | 243.41s | 221.34s | -22.07s (-9.1%) |
| target size | 7.4 GiB | 6.6 GiB | about -0.8 GiB |
| compiler artifacts | 749 | 649 | -100 (-13.4%) |

The compiler-artifact gate found all six P1 dependencies before the change and none afterward:

- `criterion`
- `agentenv-test-support`
- `aws-sdk-s3`
- `aws-config`
- `testcontainers`
- `testcontainers-modules`

The ARM64 timings are screening data rather than a replacement for the issue's x86_64 baseline; the dependency isolation result itself comes from Cargo's compiled artifact stream and dependency tree.

## Verification

- `cargo test -p agentenv --lib --locked` — 633 passed, 0 failed, 2 ignored
- `cargo test -p agentenv-e2e-tests --test snapshot_oss_e2e_test --no-run --locked`
- `cargo test -p agentenv-benchmarks --benches --no-run --locked` — 4 executables
- `cargo clippy -p agentenv -p agentenv-benchmarks -p agentenv-e2e-tests --all-targets --locked -- -D warnings`
- root dependency-tree gate — 0 forbidden dependencies
- Makefile dry-run for benchmark and OSS e2e entry points

## Not run

- Docker/MinIO-backed ignored e2e tests at runtime
- KVM/capability-dependent benchmark execution


LSX-s-Software (commented):
Thanks for contributing. Could you please fix the comment posted by the AI reviewer?

NickNYU:
> Thanks for contributing. Could you please fix the comment posted by the AI reviewer?

sure, applied the suggestion and thank you for reviewing

LSX-s-Software (approved):

PR #45, opened 2026-07-28
fix(sandbox): roll back resources after start failures

### LINKED ISSUE #42: Direct Firecracker start failures can retain sandbox resources
## Problem

Two Firecracker startup failure paths can retain resources:

1. The public `FirecrackerSandbox::start()` returns when `start_nowait()` or readiness fails without awaiting `stop()`.
2. Fresh startup applies fallible egress-policy configuration before storing the allocated network slot on the sandbox.

Orchestrator-managed launch paths already perform explicit cleanup. This issue is limited to direct/public starts and the network-policy failure window before ownership is recorded.

```text
A. Public start transaction

start_nowait
  +-- error --> stop --> return original error
  `-- ok ----> wait_for_ready
                 +-- error --> stop --> return original error
                 `-- ok ----> state: Running

B. Network slot ordering

before: allocate -> set policy -> store ownership
                       |
                       `-- error -> manager slot remains allocated

after:  allocate -> store ownership -> set policy
                                      +-- error -> stop/Drop releases slot
                                      `-- ok ---> continue boot
```

## Current behavior

Direct callers of `start()` can leave initialized runtime state and daemon-owned ublk resources retained until daemon shutdown. If fresh-start egress configuration fails before `self.network_slot` is assigned, the namespace objects are dropped but the network manager does not receive the slot allocation back.

## Expected behavior

- Treat public startup and readiness as one transaction.
- Await `stop()` after either operation fails while returning the primary startup error.
- Record ownership of an allocated network slot before running fallible configuration.
- Leave existing orchestrator cleanup and daemon shutdown behavior unchanged.

## Impact

Repeated failure paths can reduce the available ublk or network-slot pool and erode node density over time. This is especially relevant to AgentENV because it repeatedly starts and resumes large numbers of short-lived Firecracker environments.

morluto:
## Problem

Direct calls to `FirecrackerSandbox::start()` returned after startup or readiness errors without awaiting cleanup. Fresh startup also applied fallible egress-policy configuration before attaching the allocated network slot to the sandbox.

Together, these paths could retain daemon-managed ublk resources or leave a network slot unavailable after a failed start.

```text
A. Public start transaction

start_nowait
  +-- error --> stop --> return original error
  `-- ok ----> wait_for_ready
                 +-- error --> stop --> return original error
                 `-- ok ----> state: Running

B. Network slot ordering

before: allocate -> set policy -> store ownership
                       |
                       `-- error -> manager slot remains allocated

after:  allocate -> store ownership -> set policy
                                      +-- error -> stop/Drop releases slot
                                      `-- ok ---> continue boot
```

## Changes

- Treat `start_nowait()` and `wait_for_ready()` as one public-start transaction.
- Await `stop()` after either operation fails.
- Preserve the primary startup error and warn if cleanup also fails.
- Attach newly allocated network slots to the sandbox before applying fallible egress policy.

Existing orchestrator cleanup, `Drop`, and daemon shutdown behavior are unchanged.

## Commit structure

- `fix(sandbox): roll back failed direct starts`
- `fix(sandbox): retain network slot on policy failure`

The commits are separate because public-start transactionality and network-slot ownership are independently reviewable and revertible.

## Validation

- The direct-start regression test failed against the previous implementation at the runtime-state ownership assertion.
- `cargo test -p agentenv --lib sandbox::firecracker::sandbox::tests::` — 14 passed.
- `cargo clippy -p agentenv --lib -- -D warnings`
- `cargo fmt --all`

The network ownership test uses the existing deterministic test-slot seam. A full base-failing `start_fresh` integration test would require privileged networking, ublk setup, and an additional injection seam, so this PR does not add production-wide mocking for that path.

Closes #42.

### DIFF
--- a/src/sandbox/firecracker/sandbox.rs
+++ b/src/sandbox/firecracker/sandbox.rs
@@ -894,6 +894,19 @@ impl FirecrackerSandbox {
         )
     }
 
+    fn configure_network_slot(
+        &mut self,
+        slot: Slot,
+        policy: Option<&SandboxNetworkPolicy>,
+    ) -> Result<()> {
+        self.network_slot = Some(slot);
+        self.network_slot
+            .as_ref()
+            .expect("network slot was just assigned")
+            .set_egress_policy(policy)
+            .context("Failed to configure sandbox egress policy")
+    }
+
     fn mmds_metadata(&self, common: &FirecrackerCommonConfig) -> MmdsMetadata {
         common
             .mmds_metadata
@@ -1216,9 +1229,7 @@ impl FirecrackerSandbox {
         // valid DNS server IP in the 8th field. See that method for format details.
         let ip_config = slot.build_ip_boot_arg();
         let netns = slot.namespace_path();
-        slot.set_egress_policy(config.common.network_policy.as_ref())
-            .context("Failed to configure sandbox egress policy")?;
-        self.network_slot = Some(slot);
+        self.configure_network_slot(slot, config.common.network_policy.as_ref())?;
         boot_args = Some(match boot_args.take() {
             Some(existing) => format!("{existing} {ip_config}"),
             None => ip_config,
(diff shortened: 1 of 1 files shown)


LSX-s-Software:
Thanks for the contribution.

We have addressed this issue in #108 as part of a larger set of related changes.

While the approach here is valid, we chose a different solution there to better align with the overall design and avoid introducing incremental complexity across multiple small fixes.

To keep the implementation consistent, we will close this PR in favor of the consolidated approach.

Thanks again for your contribution.

Issue #15, opened 2026-07-27
Add locality-based scheduling using node cache awareness


yingdi-shan:
## Problem

AgentENV should add a locality-based scheduling strategy that places new sandboxes on nodes that already hold the most relevant cache for the requested workload.

Today the distributed scheduler can choose strategies such as round-robin, but it does not account for per-node caches. That can place a sandbox on a cold node even when another node already has useful local artifacts, such as overlaybd layers, converted image commits, memory snapshot commits, or fixed snapshot artifacts.

The scheduler should prefer warm nodes when they have enough available resources. It should only place the sandbox on colder or new nodes when the warmer candidates lack sufficient CPU, memory, disk, or other required capacity.

## Proposal

Add a locality-aware scheduling strategy, for example `locality` or `cache_aware`, that:

- Filters candidate nodes by hard resource availability.
- Scores feasible nodes by cache coverage for the requested template, snapshot, or `userImage`.
- Chooses the feasible node with the highest cache-locality score.
- Falls back to colder or new nodes only when warm nodes do not have enough resources, or when no useful cache metadata is known.

The cache score can consider local availability of rootfs/image layers, overlaybd commits, memory snapshot commits, fixed snapshot artifacts, and related P2P-published artifacts where applicable.

yingdi-shan:
We have attempted to map sandbox with cache locality awareness. However, we find that a popular image can cause severe load imbalance and eventually crash that node due to OOM. Heartbeat-based resource filtering may react too late during a burst.

A simpler alternative is resource-bounded grouped round-robin.

Each group belongs to one image and one node. Requests for the same image fill its current open group. When that group reaches any configured resource limit, it is closed and the next group is assigned to the next eligible node using a global round-robin cursor.

For example, suppose nodes are selected in the order A -> B -> C, and each group is limited to 2 sandboxes.

Placement could look like:

ubuntu group 1 -> node A
python group 1 -> node B
ubuntu group 2 -> node C
python group 2 -> node A

Requests may arrive in any order. Ubuntu requests continue filling the current Ubuntu group, while Python requests fill a separate Python group. Different images are never mixed within a group.

The scheduling process is:

Find the current open group for the requested image.
Check whether adding the VM would exceed its sandbox, CPU, or memory budget.
If it fits and the node remains eligible, place it in that group.
Otherwise, create a new one on the next eligible node.
Count pending placements immediately so concurrent requests see the reserved group capacity.
[...]

PR #51, opened 2026-07-28
feat(scheduler): add snapshot artifact locality preference

shudorcl:
## What

This PR implements the first P2P-backed snapshot locality preference slice for #15.

For `POST /sandboxes`, when `templateID` is a canonical snapshot UUID in the standard hyphenated form, the Gateway derives two advisory locality requirements using the existing fixed snapshot artifact identities:

- `snapshot/v1/artifacts/{snapshot_id}/vm_state.bin`
- `snapshot/v1/artifacts/{snapshot_id}/firecracker-manifest.json`

After resource-limit filtering, the Scheduler prefers eligible nodes that advertise the highest coverage of those artifacts in the existing P2P artifact index. The configured scheduling strategy performs the final selection among tied nodes.

Template aliases remain supported, but intentionally do not produce locality hints because the Gateway has no authoritative pre-scheduling alias resolver. Alias requests therefore use the existing scheduling behavior.

## Why

The scheduler currently ignores whether an eligible node already provides the snapshot artifacts needed to start a sandbox. It can therefore select a cold node even when another feasible node advertises the required artifacts through the existing P2P index.

This change introduces a bounded, advisory locality preference without duplicating load-aware selection or changing runtime fallback behavior. It establishes the request and inventory path needed for later locality work while keeping the scheduler's artifact store limited to artifact-key-to-node mappings.

## Related issue

Refs #15

This PR is one focused part of #15 and does not close the full issue.

## Scope and non-goals

In scope:

- derive fixed snapshot artifact requirements from canonical snapshot IDs;
- carry advisory locality requirements through the internal scheduling protocol;
- prefer the eligible nodes with the highest positive artifact coverage;
- preserve existing selection behavior when locality data is absent or stale;
- bound request inspection, requirement count, and artifact lookup work;
- remove a node's artifact mappings when it is unregistered;
- document and test the new scheduling flow.

Non-goals:

- resolving template aliases before scheduling;
- resolving OCI image references into OverlayBD layer requirements;
- indexing non-P2P local caches;
- adding locality requirements for `POST /sandboxes-cold`;
- introducing a monolithic `cache_aware` strategy;
- implementing or duplicating load-aware node selection;
- changing hard resource filtering or runtime artifact fallback behavior.

The scheduling flow becomes:

```text
ready nodes
    -> hard resource-limit filtering
    -> snapshot artifact locality preference
    -> configured strategy selection
    -> selected node
```

The Gateway inspects only exact `POST /sandboxes` creation requests. It buffers at most 64 KiB while extracting `templateID` and restores the original body for upstream forwarding. Oversized, malformed, empty, alias-based, and non-canonical template IDs produce no locality requirements and continue through normal scheduling.

For a canonical snapshot UUID, the Gateway normalizes the ID and emits the two fixed artifact keys listed above. It does not query runtime nodes to resolve aliases. Such fan-out would be non-authoritative, could observe inconsistent alias mappings, would amplify unauthenticated request work, and could expand credential exposure.

The Scheduler validates the request-controlled locality requirement count before node discovery. After resource filtering, `PreferLocalNodes`:

- trims and deduplicates requirement keys;
- isolates lookups by cluster ID and P2P backend;
- scores only already-eligible candidates;
- retains all nodes tied for the highest positive coverage;
- preserves the complete eligible set when no useful locality data exists.

Artifact lookup intersects provider records with the eligible node-ID set inside the store, so scheduling work scales with current candidates rather than all historical providers. Locality does not create reservations or other request-lifetime state.

Locality remains advisory. Stale or missing records do not fail sandbox creation, and the selected runtime continues through its normal local cache, P2P, snapshot repository, or registry fallback path.

- Public API or generated protocol: Adds optional `ScheduleRequestHint.locality_requirements`, `LocalityRequirement.key`, and `NewSandboxHint.template_id` fields to the internal scheduler protobuf. The additions are wire-compatible. Older senders omit them, and receivers without locality support ignore the unknown fields.
- Configuration or defaults: No configuration fields or defaults change. Existing `round_robin` and `random` strategy behavior is unchanged when locality requirements or provider records are unavailable.
- Snapshot manifest, artifact layout, or storage format: No format or layout changes. The implementation reuses the existing `snapshot/v1/artifacts/{snapshot_id}/...` identities.
- Upgrade and rollback: No persistent migration is required. Gateway and Scheduler versions can be rolled back without changing snapshot data or the artifact index; rollback removes the placement preference and restores the previous selection behavior.
- Host requirements, permissions, ports, or dependencies: No new host permissions, services, ports, or runtime packages. `github.com/google/uuid` is promoted from an indirect to a direct Go dependency at the same version.

## Validation

- [ ] `make fmt`
- [ ] `make clippy`
- [ ] `make test-unit`
- [x] `make -C services test` (required when `services/` changes)
- [x] Documentation updated

Commands and results:

```text
make -C services proto PROTOC=<protoc 3.21.12>
gofmt -w services/api/proto/scheduler.pb.go services/api/proto/scheduler_grpc.pb.go
# PASS: protoc-gen-go 1.36.11 and protoc-gen-go-grpc 1.6.2; no tracked diff

make -C services build
make -C services test
make -C services fmt-check
make -C services vet
# PASS with Go 1.25.0

cd services
go test -race ./gateway/internal ./scheduler/internal
# PASS with Go 1.25.0

git diff --check
# PASS
```

The Services commands above were run locally. `fmt-check` was verified from an LF-normalized export of the same commit because this WSL checkout uses `core.autocrlf=true`.

The following remote Services CI and integration jobs also passed for commit `b2f9346577b001bba28fb038d46ae0beb233c956`:

- Services CI: https://github.com/shudorcl/AgentENV/actions/runs/30394461927
- Integration Tests: https://github.com/shudorcl/AgentENV/actions/runs/30394461784

Skipped checks and reasons:

- Root Rust formatting, Clippy, and unit-test targets were not run because this PR changes only the Go control-plane module and its generated Go protobuf. The remote AgentENV integration, single-node E2E, and docker-compose E2E jobs passed.
- No benchmark was run. The scheduling hot path is bounded to at most 32 locality requirements, and provider lookup is intersected with the current eligible candidate set. The PR adds focused boundary and stale-record tests, but does not claim a measured performance improvement.

## Risks and reviewer notes

- Locality metadata is eventually consistent and can be stale. The preference is deliberately advisory, and no positive match preserves the complete eligible candidate set.
- Requests with aliases receive no locality optimization. Supporting aliases safely requires an authoritative control-plane resolver and is left for a follow-up.
- Hint extraction occurs before upstream authentication. Request-body inspection is therefore bounded to 64 KiB and does not fan out to runtime nodes.
- The locality requirement list is request-controlled. The Scheduler rejects more than 32 requirements before node snapshotting or enrichment.
- Popular artifacts may have many historical providers. `LookupEligible` bounds work to the current eligible candidate IDs, and unregistering a node removes its artifact mappings.
- Reviewers should start with `services/gateway/internal/schedule_hint.go`, `services/gateway/internal/template_resolution.go`, `services/scheduler/internal/locality.go`, and `services/scheduler/internal/service.go`.

## Checklist

### DIFF
--- a/services/README.md
+++ b/services/README.md
@@ -113,7 +113,10 @@ The strategy interface receives `RichNode` values that carry the node identity (
 ### Snapshot locality preference
 
 For `POST /sandboxes`, the Gateway derives advisory locality requirements from
-the requested `templateID` using the snapshot's stable P2P artifact keys:
+the requested `templateID` using the snapshot's stable P2P artifact keys. UUID
+template IDs are normalized to their canonical lowercase form. When the
+request uses a template alias, the Gateway resolves it through the discovered
+nodes' alias endpoint before building the keys:
 
 - `snapshot/v1/artifacts/{snapshot_id}/vm_state.bin`
 - `snapshot/v1/artifacts/{snapshot_id}/firecracker-manifest.json`
@@ -125,11 +128,12 @@ that preferred set. Artifact lookups use each candidate node's heartbeat
 reported cluster ID and P2P backend, so records do not cross cluster or backend
 boundaries.
 
-Locality is best-effort. If the request has no requirements, no eligible node
-has a matching record, or a node has not reported P2P context, the scheduler
-passes the full eligible candidate set to the configured strategy. Stale
-artifact records therefore affect placement only; the selected AgentENV node
-continues to use its normal local cache, P2P, repository, or registry fallback.
+Locality is best-effort. If alias resolution is unavailable, the request has no
+requirements, no eligible node has a matching record, or a node has not
+reported P2P context, the scheduler passes the full eligible candidate set to
+the configured strategy. Stale artifact records therefore affect placement
+only; the selected AgentENV node continues to use its normal local cache, P2P,
+repository, or registry fallback.
 
 This initial implementation does not resolve OCI image references into
 OverlayBD layer requirements and does not index non-P2P local caches.
--- a/services/README.md
+++ b/services/README.md
@@ -114,9 +114,8 @@ The strategy interface receives `RichNode` values that carry the node identity (
 
 For `POST /sandboxes`, the Gateway derives advisory locality requirements from
 the requested `templateID` using the snapshot's stable P2P artifact keys. UUID
-template IDs are normalized to their canonical lowercase form. When the
-request uses a template alias, the Gateway resolves it through the discovered
-nodes' alias endpoint before building the keys:
+template IDs must use the standard hyphenated form and are normalized to
+lowercase before building the keys:
 
 - `snapshot/v1/artifacts/{snapshot_id}/vm_state.bin`
 - `snapshot/v1/artifacts/{snapshot_id}/firecracker-manifest.json`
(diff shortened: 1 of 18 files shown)


shudorcl:
I initially addressed alias-aware locality through gateway-to-node fan-out, but follow-up review showed that this approach was neither authoritative nor safe with respect to credential exposure. This PR now intentionally limits locality hints to canonical snapshot IDs; alias-based sandbox creation remains fully supported and falls back to normal scheduling.
Maybe a follow-up can add an authoritative control-plane alias resolver and use it to support alias-aware locality without node fan-out.

yingdi-shan:
Thank you for your contribution! I've updated a comment in #15. We tried this approach previously; however, we found that a popular image can cause severe load imbalance and eventually crash the node due to OOM. Additionally, since we overcommit memory heavily, we don't know in advance how much memory each sandbox will consume, which makes it difficult to schedule them properly.

shudorcl:
Thanks for the additional context. I see how #51 could lead to the load imbalance you described. Since the resource-bounded grouped round-robin approach is quite different, I'll close this PR and start fresh with a new one.

PR #56, opened 2026-07-29
fix(proxy): isolate connections across runtime generations

MasakiMu319:
## Problem

Sandbox interaction IPs are recycled when runtimes stop, while envd always listens on the same port. The process-wide Hyper client pools idle connections by URI authority (`IP:port`), which does not encode the sandbox runtime generation.

This allows the following sequence:

1. A request to runtime A leaves an idle keep-alive connection in the pool.
2. Runtime A stops and its interaction IP is assigned to runtime B.
3. The next request for runtime B reuses A's stale TCP flow.
4. The new guest rejects that flow with an RST and Hyper returns `client error (SendRequest)`; the proxy returns 502.

Retrying automatically is unsafe because proxied ConnectRPC requests can have side effects.

## Fix

- Disable idle per-host pooling for the server-to-sandbox proxy client with `pool_max_idle_per_host(0)`.
- Add a behavioral regression test that serves one request over a keep-alive connection, simulates the address moving to a new runtime generation, and verifies that the second request opens a fresh TCP connection.

Active requests and streaming connections are unchanged; only reuse after a connection becomes idle is disabled.

## Validation

The regression test was run against the current `main` before the production change and failed with:

```text
proxy client reused an idle TCP connection after the runtime generation changed
```

After the fix:

- `cargo test -p agentenv --lib api::proxy::tests` — 32 passed
- focused regression test repeated 50 times — 50 passed
- `cargo clippy -p agentenv --all-targets -- -D warnings`

A same-host lifecycle probe reproduced 15 initial proxy failures in 200 cycles before the fix and 0 in 1,200 cycles afterward, with no retries. Median latency remained effectively unchanged:

- create: 117.18 ms → 117.25 ms
- first exec: 81.26 ms → 81.46 ms

### DIFF
--- a/src/api/proxy.rs
+++ b/src/api/proxy.rs
@@ -120,7 +120,11 @@ fn auto_resume_min_sandbox_timeout() -> Duration {
 pub(crate) fn build_proxy_client() -> ProxyClient {
     let mut connector = HttpConnector::new();
     connector.set_connect_timeout(Some(PROXY_CONNECT_TIMEOUT));
-    Client::builder(TokioExecutor::new()).build(connector)
+    // Interaction IPs are reused across sandbox runtime generations. Hyper keys
+    // its idle pool by authority, so a pooled connection can retain a stale VM flow.
+    Client::builder(TokioExecutor::new())
+        .pool_max_idle_per_host(0)
+        .build(connector)
 }
 
 pub(crate) fn router<I>(api_impl: I) -> Router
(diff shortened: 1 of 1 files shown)


LSX-s-Software (commented):
Good fix. However, the mapping from "runtime generation change" to the two socket paths is not immediately obvious. Could you add a short comment above `simulate_runtime_generation_change` explaining that bytes on `previous_runtime` mean stale connection reuse, while a new `accept()` represents connecting to the next generation? It would also help to document that the returned boolean means "previous connection was reused".

MasakiMu319:
@LSX-s-Software Thanks — clarified the socket-to-runtime mapping and the boolean return value in commit [`9a3bae2`](https://github.com/kvcache-ai/AgentENV/pull/56/commits/9a3bae2f395e1fe9c6424c910280d21f8b60bfc4).

LSX-s-Software (approved):

PR #64, opened 2026-07-29
fix(sandbox): resume sandbox upon recoverable pause failure

### LINKED ISSUE #41: Recoverable pause failures can leave sandboxes frozen while reported as Running
## Problem

A recoverable backend pause failure can occur after Firecracker has already been paused. The orchestrator currently restores the sandbox handle, proxy route, and `Running` state without first calling `resume()`.

The sandbox can therefore be discoverable and routable as `Running` while its VM remains frozen.

```text
pause() returns recoverable error
                |
                v
             resume()
                |
                +-- success
                |      |
                |      v
                |   restore handle + route
                |      |
                |      v
                |   state: Running
                |      |
                |      v
                |   return pause error
                |
                `-- failure
                       |
                       v
                    stop + remove
                       |
                       v
                    state: Absent
                       |
                       v
                    return combined error
```

## Current behavior

On a non-terminal `SandboxPauseError`, `pause_sandbox_inner` reinserts the detached handle, restores the route, and transitions `Pausing -> Running`.

The backend contract only says that a recoverable failure leaves the runtime safe to resume; it does not guarantee that the VM is already running.

## Expected behavior

- Call the backend's `resume()` before restoring the handle, route, and `Running` state.
- Restore the sandbox only after resume succeeds.
- If resume also fails, stop and remove the sandbox instead of publishing a false healthy state.
- Preserve useful context from both the pause and recovery failures.

## Impact

Pause and resume are central to AgentENV's low-cost idle environment model. Returning a frozen VM to the running pool can leave agent workloads apparently healthy but unable to make progress, while routing and resource metrics continue to treat the sandbox as active.

LSX-s-Software:
## What

Ensure `SandboxBackend::pause()` returns a recoverable error only after the sandbox has been restored to a running state.

For the Firecracker backend, a recoverable pause failure now triggers an in-place resume before the error is returned. If recovery also fails, the error is promoted to terminal so the orchestrator stops and removes the sandbox instead of publishing a frozen VM as `Running`.

## Why

Firecracker is paused before snapshot artifacts are fully created. A later snapshot-processing failure could therefore leave the VM paused while the orchestrator restored its handle, proxy route, and `Running` metadata. The sandbox appeared healthy and routable but could not make progress.

Keeping the recovery guarantee at the backend boundary lets the orchestrator retain its simple error handling without depending on backend-specific intermediate states.

## Related issue

Closes #41

## Scope and non-goals

Included:

- Define the recoverable `pause()` contract at the sandbox backend boundary.
- Resume Firecracker after recoverable pause failures.
- Promote failed recovery to a terminal error with both failure causes.
- Keep the mock backend aligned with the production contract.
- Add focused orchestrator regression coverage.

Not included:

- Lifecycle concurrency or delete-path refactoring.
- Changes to the low-level `FirecrackerSandbox::pause_to_dir()` semantics.
- New public APIs, configuration, or persisted state.

```text
backend pause failure
        |
        +-- terminal ----------------------> return Terminal
        |
        `-- recoverable --> resume backend
                              |
                              +-- success --> return original Recoverable
                              |
                              `-- failure --> return Terminal(pause + resume errors)
```

The orchestrator can safely restore the handle, proxy route, and `Running` metadata for a recoverable error because the backend now guarantees that recovery has already completed.

- Public API or generated protocol: N/A; no API or generated-code changes.
- Upgrade and rollback: No migration required; rollback restores the previous pause failure behavior.

## Validation

- [x] `make fmt` (equivalent command run directly)
- [x] `make clippy` (equivalent focused command run directly)
- [x] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
cargo test -p agentenv --lib pause_ -- --nocapture
# 17 passed; 0 failed

cargo fmt --all -- --check
# passed

cargo clippy -p agentenv --all-targets -- -D warnings
# passed
```

Skipped checks and reasons:

- Services tests were not run because `services/` is unchanged.
- Code generation and benchmarks are not applicable.

## Risks and reviewer notes

N/A

## Checklist

### DIFF
--- a/src/orchestrator/tests.rs
+++ b/src/orchestrator/tests.rs
@@ -2808,6 +2808,16 @@ async fn orchestrator_delete_paused_sandbox_removes_metadata() -> Result<()> {
 async fn pause_failure_rolls_back_to_running_and_preserves_handle() -> Result<()> {
     setup();
     let behavior = Arc::new(MockBehavior::new());
+    let resume_calls = Arc::new(StdMutex::new(0usize));
+    let resume_calls_for_hook = Arc::clone(&resume_calls);
+    behavior.set_on_operation(
+        MockOperation::Resume,
+        Arc::new(move || {
+            *resume_calls_for_hook
+                .lock()
+                .expect("resume call counter mutex poisoned") += 1;
+        }),
+    );
     behavior.push_action(
         MockOperation::Pause,
         MockAction::Fail {
@@ -2815,7 +2825,8 @@ async fn pause_failure_rolls_back_to_running_and_preserves_handle() -> Result<()
         },
     );
     let orchestrator =
-        make_orchestrator_with_factory(MockBackendFactory::with_behavior(behavior)).await;
+        make_orchestrator_with_factory(MockBackendFactory::with_behavior(Arc::clone(&behavior)))
+            .await;
 
     let created = orchestrator
         .create_sandbox(create_request(Some(60), &[("team", "pause-failure")]))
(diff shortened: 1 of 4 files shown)


yingdi-shan (approved):
LGTM.

PR #80, opened 2026-07-29
fix(orchestrator): fail fork when source cannot leave Forking

Ayush7614:
## Summary
- After a successful backend fork, restoring the source from `Forking` → `Running` was best-effort and ignored on failure.
- That left the source stuck in `Forking` while children were still registered, breaking later pause/resume/delete waiters.
- Fork now fails hard on that restore error, stops forked children, and surfaces the store failure.

## Test plan
- [x] Added `fork_sandbox_fails_when_source_cannot_leave_forking`
- [x] `cargo test -p agentenv --lib fork_sandbox_`

### DIFF
--- a/src/orchestrator/service.rs
+++ b/src/orchestrator/service.rs
@@ -561,7 +561,10 @@ where
             }
         };
 
-        // Restore the source sandbox's state to Running.
+        // Restore the source sandbox's state to Running. A failure here must
+        // abort the fork: leaving the source in Forking blocks pause/resume/
+        // delete waiters, and registering children would report success while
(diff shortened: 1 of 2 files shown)


LSX-s-Software:
Thanks for the contribution.

This change targets a scenario that is theoretically possible but practically negligible. In contrast, the fix introduces non-trivial complexity and ongoing runtime and maintenance cost.

We generally avoid adding complexity for edge cases that are unlikely to happen and have minimal real-world impact.

Given this trade-off, we are not planning to merge this PR.

Appreciate your contribution.

PR #81, opened 2026-07-29
fix(orchestrator): roll back network policy when metadata persist fails

Ayush7614:
## Summary
- Network policy updates applied live iptables/runtime rules before persisting metadata.
- A store failure returned an error while egress rules had already changed, desyncing API metadata from enforced policy.
- On persist failure, roll the runtime policy back to the previous value and map `StateConflict` to 409 like patch-params.

## Test plan
- [x] Added `sandbox_network_policy_rolls_back_when_metadata_persist_fails`
- [x] `cargo test -p agentenv --lib sandbox_network_policy`

### DIFF
--- a/src/orchestrator/service.rs
+++ b/src/orchestrator/service.rs
@@ -1427,10 +1427,38 @@ where
 
     #[tracing::instrument(skip(self, network_policy), fields(sandbox_id = %sandbox_id))]
     pub async fn replace_sandbox_network_policy(
+        self: &Arc<Self>,
+        sandbox_id: SandboxId,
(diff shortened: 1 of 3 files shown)


LSX-s-Software:
Thanks for the contribution.

We have addressed this issue in #108 as part of a larger set of related changes.

While the approach here is valid, we chose a different solution there to better align with the overall design and avoid introducing incremental complexity across multiple small fixes.

To keep the implementation consistent, we will close this PR in favor of the consolidated approach.

Thanks again for your contribution.

PR #82, opened 2026-07-29
fix(orchestrator): treat persistence cleanup failures as operation failures

Ayush7614:
## Summary
- Delete reported success even when paused artifacts could not be removed, so sandboxes could resurrect after restart.
- Resume warned and continued when clearing the `Resuming` marker failed, so a crash could discard the paused record.
- Delete now fails and restores a stable state for retry; resume rolls back to `Paused` if record cleanup fails.

## Test plan
- [x] Added `delete_sandbox_fails_when_persisted_state_cannot_be_removed`
- [x] Added `resume_delete_record_failure_rolls_back_to_paused`
- [x] `cargo test -p agentenv --lib resume_marks_resuming`
- [x] `cargo test -p agentenv --lib delete_sandbox_fails_when_persisted`

### DIFF
--- a/src/orchestrator/persistence/file_backed.rs
+++ b/src/orchestrator/persistence/file_backed.rs
@@ -365,8 +365,10 @@ impl SandboxPersister for FileBackedSandboxPersister {
 
     async fn delete_record_and_artifacts(&self, sandbox_id: &SandboxId) -> PersistenceResult<()> {
         debug!(sandbox_id = %sandbox_id, "deleting paused sandbox record and artifacts");
-        self.remove_record(sandbox_id).await?;
+        // Remove artifacts before the record so a failed artifact cleanup cannot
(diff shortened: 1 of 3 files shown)


LSX-s-Software:
Thanks for the contribution. After reviewing this, I don’t think we should merge this change.
The PR is based on injected persistence-cleanup failures, but it does not provide a real-world reproduction or demonstrate a user-facing issue under normal operation. These failures indicate an underlying local storage problem, and the proposed rollback and one-off retry logic cannot make the lifecycle operation truly transactional. It also adds substantial state-machine complexity and may unnecessarily fail or roll back an otherwise successful delete or resume operation.
The existing best-effort cleanup with explicit warnings is the intended behavior here. Closing this PR as the added complexity is not justified by a concrete production failure.

PR #85, opened 2026-07-30
feat(overlaybd): introduce index() and flatten_with_args() from LSMTFile

huang-jl:
## What

Expose two new capabilities from the read-write LSMTFile:

  - Add index() to return the shared Arc<RwLock<ComboIndex>>.
  - Add flatten_with_args() to support custom CommitArgs.

The existing flatten() method remains backward-compatible and delegates to flatten_with_args() with default arguments.

## Why

Higher-level callers need access to the combined LSMT index and finer control over flatten operations, including the destination writer, UUID metadata, user tags, and write concurrency.

Previously, these capabilities were only available internally.

index() clones and returns the internal Arc<RwLock<ComboIndex>>. Callers share the same index instance with LSMTFile and must use the existing lock for synchronization.

flatten_with_args():

  1. Loads the current virtual size.
  2. Acquires an index read lock and resolves mappings for the full logical address range.
  3. Releases the index lock.
  4. Passes the mappings, source layers, virtual size, and caller-provided CommitArgs to compact_to().

The existing method now delegates as follows: `flatten_with_args(CommitArgs::new(dest_file))`.

Existing callers therefore retain the same default metadata and sequential write behavior. Read and write failures continue to propagate from compact_to().

- Public API or generated protocol: Adds public Rust APIs to the OverlayBD crate. No HTTP or generated protocol changes.
- Configuration or defaults: No changes.
- Snapshot manifest, artifact layout, or storage format: No default behavior or format changes.
- Upgrade and rollback: No migration is required. Code using the new methods will no longer compile after rollback.
- Host requirements, permissions, ports, or dependencies: No changes.

## Checklist

### DIFF
--- a/storage/overlaybd/src/lsmt/file/readwrite.rs
+++ b/storage/overlaybd/src/lsmt/file/readwrite.rs
@@ -58,6 +58,10 @@ pub struct LSMTFile {
 }
 
 impl LSMTFile {
+    pub fn index(&self) -> Arc<RwLock<ComboIndex>> {
+        self.index.clone()
+    }
+
     fn rw_layout_from_header(header: &HeaderTrailer) -> Result<RwLayout> {
         let is_sparse_rw = header.is_sparse_rw();
         let is_hybrid_rw = header.is_hybrid_rw();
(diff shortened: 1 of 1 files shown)


yingdi-shan (approved):
LGTM.

Issue #37, opened 2026-07-28
Add dynamic shell completion to the aenv CLI


LSX-s-Software:
## Summary

Add first-class shell completion support to the native `aenv` CLI for Bash, Zsh, and Fish.

In addition to completing commands and flags, the CLI should dynamically complete AgentENV resources such as sandbox IDs and template/snapshot names.

## Motivation

The CLI contains nested commands, aliases, and resource identifiers that are difficult to enter manually:

```bash
aenv snapshot create <sandbox-id>
aenv template watch <template>
aenv resume <sandbox-id>
```

Sandbox IDs are UUIDs, while templates and snapshots may be referenced by either names or IDs. As the number of resources grows, users currently need to run a list command, copy a value, and paste it into the next command.

Shell completion would make interactive CLI use faster and reduce invalid commands caused by stale or mistyped identifiers.

## Proposed interface

Add a command that emits completion registration code:

```bash
aenv completion bash
aenv completion zsh
aenv completion fish
```

The generated registration should call back into the current `aenv` binary when completion is requested, allowing it to provide both static and dynamic candidates.

Static completion should cover:

- top-level and nested subcommands;
- command aliases such as `ls`, `rm`, `cn`, and `snap`;
- flags and option values;
- enum values such as `--output table|json`;
- local paths such as the Dockerfile argument to `aenv build`.

### Sandboxes

Commands accepting a sandbox ID should query the sandbox list and filter candidates by the states accepted by the command:

- `pause`, `exec`, `timeout`, and `snapshot create`: running sandboxes;
- `resume`: paused sandboxes;
- `connect`, `delete`, and `snapshot list --sandbox-id`: running and paused sandboxes.

The inserted value should be the sandbox ID. When supported by the shell, the candidate description should include the sandbox's template and state.

### Templates and snapshots

Commands accepting a template or snapshot should complete human-readable names or aliases first, with IDs available as fallback candidates:

- `aenv start <target>`;
- `aenv template watch <template>`;
- `aenv template delete <template>`.

`aenv start --cold` should not query template or snapshot candidates because its target is an OCI image reference.

Candidate descriptions should distinguish templates from snapshots and include the underlying ID and build status where available.

## Failure and performance behavior

Dynamic completion is an interactive, best-effort feature and must not make the shell feel blocked.

- Use short connection and request timeouts.
- Limit and deduplicate returned candidates.
- Filter candidates by the prefix already entered.
[...]

PR #89, opened 2026-07-30
feat(cli): add shell completion scaffolding

rbalachandar:
Splits #37 into a static-first scaffolding change. This adds the `aenv completion <shell>` command and the static completion surface; dynamic resource completion and automatic installer integration are deferred to follow-up work.

## What it does

- `aenv completion bash|zsh|fish` emits a static completion script via `clap_complete`. The command tree is rebuilt from the `Cli` derive spec (`crate::Cli::command()`), so the generated script always reflects the real CLI. There is no separate completion definition to keep in sync.
- Covers everything #37 lists under static completion: top-level and nested subcommands, command aliases, flags, the `--output table|json` enum, and local path arguments such as the Dockerfile passed to `aenv build`.
- Documents manual installation into standard Bash, Zsh, and Fish completion directories, including the Zsh `fpath` guidance and one-session activation commands.
- Elvish and PowerShell are intentionally excluded; #37 scopes the shells to Bash, Zsh, and Fish.

## Behavior change worth calling out

The command aliases `cn`, `ls`, `rm`, `snap`, `templates` and the `snapshot`/`template` subcommand aliases use `visible_alias`. `clap_complete` emits only visible aliases, so this is required for alias completion to work. The visible side effect is that these aliases now also appear in `--help`. Runtime parsing is unchanged; the aliases were already accepted as input.

## Dependency

Adds `clap_complete = "4"`, matching the existing `clap = "4"` major.

## Testing

- `cargo test -p aenv --bin aenv commands::completion` — 11 completion tests covering Bash, Zsh, Fish, aliases, the `--output` enum, nested subcommands, and output/flush error handling.
- `cargo fmt --all -- --check` passes.
- `cargo check -p aenv --bin aenv` passes.

The workspace-wide `make fmt` / `make clippy` also build the server and storage crates, which require Linux and `/dev/kvm`, so only the `aenv`-scoped checks were run here. No integration tests apply to this CLI-only change.

## Out of scope

Dynamic resource completion (sandbox IDs filtered by command-applicable state, template/snapshot names, and the `aenv start --cold` OCI-ref exception) and automatic completion installation/removal. The installer work will be handled separately using standard completion directories without modifying shell startup files.

Refs #37 - static scaffolding portion; dynamic completion and installer integration remain separate follow-up work.


crates/aenv/src/commands/completion.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/// Generate a static completion script for the requested shell and write it to
+/// stdout. The command tree is rebuilt from the live `Cli` derive spec via
+/// `crate::Cli::command()` so completion can never drift from the real CLI.
+pub fn run(args: Args) -> Result<()> {
+    let mut cmd = crate::Cli::command();
+    let mut out = std::io::stdout();
+    clap_complete::generate(ClapShell::from(args.shell), &mut cmd, "aenv", &mut out);
+    out.flush()?;
```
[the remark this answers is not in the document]
rbalachandar: Applied suggestion

crates/aenv/src/commands/completion.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn includes_subcommands() {
+        // Subcommand groups flow through from the derive spec: `snapshot`
+        // exposes a `create` child, so both must appear in the script. (`create`
+        // is unique to `snapshot create` — no top-level command uses it.)
+        let s = generate_for(Shell::Bash);
+        assert!(
+            s.contains("snapshot") && s.contains("create"),
+            "completion should reference the `snapshot` command and its `create` \
+             subcommand; got:\n{s}"
+        );
```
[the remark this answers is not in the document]
rbalachandar: Addressed in b200901. Swapped the loose substrings for bash-specific fragments that encode the  context — aenv,cn) (cn dispatched at top level), aenv__subcmd__snapshot__subcmd__create (snapshot→create key), and compgen -W "table json" (the --output enum word list). Each only exists when the feature is present.

crates/aenv/src/commands/completion.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    #[test]
+    fn includes_subcommands() {
+        // This hierarchical dispatch key exists only when `create` is emitted as
+        // a child of `snapshot`.
+        let s = generate_for(Shell::Bash);
+        assert!(
+            s.contains("aenv__subcmd__snapshot__subcmd__create"),
+            "completion should register `snapshot create` as a subcommand; got:\n{s}"
+        );
```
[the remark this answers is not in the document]
rbalachandar: Addressed in 9a3b38e. All three feature checks now go through Cli::command() metadata instead of  generated-script internals: connect's visible aliases, the snapshot → create subcommand, and --output's   possible values (get_possible_values). Only the per-shell registration smoke tests remain on the  generated output, as suggested.

LSX-s-Software (commented):
Thanks for contributing. Could you please add some documentation to `docs/src/getting-started/aenv-cli.md` explaining how to use the completion command? In addition, could you help us implement the automatic shell completion installation mentioned in the issue?

rbalachandar:
> Thanks for contributing. Could you please add some documentation to `docs/src/getting-started/aenv-cli.md` explaining how to use the completion command? In addition, could you help us implement the automatic shell completion installation mentioned in the issue?

Added the documentation. Related to dynamic completion added a comment https://github.com/kvcache-ai/AgentENV/issues/37#issuecomment-5135724573 related to clap . Could you please have a look and let know the best option, based on that I will do follow-up PR for dynamic shell completion.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,48 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
+aenv completion bash
+aenv completion zsh
+aenv completion fish
+```
+
+Each command writes the matching script to stdout, so redirect it wherever your shell expects:
+
+```bash
+aenv completion bash > ~/.local/share/aenv-completion.bash
+aenv completion zsh  > ~/.config/aenv/_aenv
```
LSX-s-Software: These two paths seem incorrect. Could you please update them based on bash and zsh's document?

LSX-s-Software (changes requested):
Thanks for adding the completion documentation. However, the documented installation paths are not correct for Bash and Zsh. Could you please update these instructions based on the Bash and Zsh documentation?

It would also be preferable to implement the automatic installation and removal requested in issue #37: `make install-aenv`, `scripts/install-cli.sh`, and the full installer should install lightweight completion loaders into the standard shell completion directories, while `make uninstall-aenv` should remove them. The loaders should generate completion code from the currently installed `aenv` binary when the shell starts, so upgrading the CLI does not leave stale generated scripts behind.

Regarding the dynamic completion feature, I think it's OK to wait for `clap` to support it in a stable version.

rbalachandar:
## Shell-completion installers — status and dispositions

Heads-up for reviewers on the `aenv` shell-completion installer work, since the OpenCodeReview bot re-runs on every push and re-flags the same declined items each round (it doesn't track replies/resolutions). This comment lists what's implemented, what's intentionally declined, and why, so the human review isn't obscured by the recurring bot noise.

### Implemented (per the issue #37 ask + hardening)
- Regenerating loaders installed/removed by `make install-aenv`/`uninstall-aenv`, `scripts/install-cli.sh`, and `scripts/install.sh`: bash (lazy), fish (startup), zsh user (rc-snippet), zsh system (static `_aenv`). Completion is always generated from the installed `aenv`, so it never goes stale across upgrades.
- All writes go through one atomic-commit primitive (same-dir temp + rename, symlink-resolved, mode/owner preserved, empty/partial output rejected).
- Portable across GNU/Linux and BSD/macOS (no `readlink -f`, no `chown --reference` dependency); idempotent; resists malformed/partial state without truncating the user's rc; ownership marker so install/uninstall won't clobber a hand-written or package-managed completion; `set -u`/unset-`HOME` safe; failing/empty `aenv completion` never damages an existing file.

### Intentionally declined (with rationale)
These recur on every bot run; I'm recording the rationale once rather than re-litigating each round:
1. **Privilege routing for system-mode writes from `install-cli.sh`.** A `curl … | bash` installer silently invoking `sudo` mid-stream is surprising and risky UX. The Makefile path (`AENV_INSTALL_SUDO`) and `install.sh` (run as root) install system completions correctly; the rare `install-cli.sh`-into-`/usr/local` case degrades to a clear warning.
2. **File-level locking for concurrent installs.** Concurrent installs into the same `\$HOME` are not a realistic target for a CLI installer, and the atomic-rename already prevents partial/corrupt state (the only real data-loss vector). A lock would add a privileged-run symlink surface and a leftover artifact for no realistic gain.
3. **Ownership/content gate before replacing completion stubs.** The destination filenames are aenv-specific (`completions/aenv`, `aenv.fish`, `_aenv`). A content gate was nonetheless added (install + uninstall now both check the ownership marker before touching an existing file).
4. **`sed` byte-exactness in the drift checker.** The three copies are extracted and compared on the same platform in one run; the cross-platform EOF-newline concern is hypothetical and guarded by the blocks always terminating with a newline.

Happy to revisit any of these if a maintainer disagrees — flagging them here so the recurring bot comments can be read as known/decided rather than new.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,65 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
 
 To delete a snapshot, use `aenv template delete <snapshot-id>` or `aenv template delete <name>` — snapshots share the same underlying store as templates and are deleted through the same command.
+
+---
+
+## Shell completion
+
+`aenv completion <shell>` prints a shell-completion script to stdout. The script is rebuilt from the live CLI definition on every run, so it always matches the installed `aenv` — top-level commands, nested subcommands (e.g. `aenv snapshot create`), visible aliases (`cn`, `ls`, `rm`, `snap`, `templates`), flags, the `--output table|json` enum, and local path arguments such as the Dockerfile passed to `aenv build`.
+
+Three shells are supported; elvish and powershell are explicit non-goals.
```
LSX-s-Software: Please remove these lines.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,65 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
+
+### Installed automatically
+
+The installers set up **regenerating** completion loaders for you, so the completion is always generated from the currently installed `aenv` and never goes stale across upgrades:
+
+- `make install-aenv` / `make uninstall-aenv` — installs/removes the loaders (set `AENV_INSTALL_COMPLETION=0` to skip). User-local installs (`AENV_INSTALL_PREFIX=~/.local`) write per-user loaders and an `~/.zshrc` snippet; system installs write under `<prefix>/share`.
+- `scripts/install-cli.sh` — installs loaders into the matching user-local or system directories based on the install location.
+- `scripts/install.sh` (full installer) — installs system-wide loaders under `/usr/local/share`.
+
+After install, open a new shell (or re-source your rc) and the loaders take effect. If you installed with one of the methods above, you can skip the manual steps below — they remain as a fallback for users who skipped the installer or want to customize the location.
```
LSX-s-Software: Please reword and simplify them.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,65 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
+
+For a one-session test, evaluate the script in the current shell:
+
+```bash
+source <(aenv completion bash)        # bash
+eval "$(aenv completion zsh)"         # zsh
+aenv completion fish | source         # fish
+```
+
+To make completion persistent, add the matching line to your shell's rc file (`~/.bashrc` or `~/.bash_profile`, `~/.zshrc`, `~/.config/fish/config.fish`) and restart the shell or re-source the file.
```
LSX-s-Software: Please remove this line.

LSX-s-Software (changes requested):
Thanks for implementing the automatic completion installation as requested. However, I did not expect the installation support to require this much new infrastructure. After reviewing the resulting implementation, I have two blocking design concerns.

## 1. The installer must not modify `~/.zshrc`

The user-mode installation currently installs Zsh completion by inserting a managed block into `~/.zshrc`.

This directly conflicts with issue #37, which explicitly states:

> Installers should use standard completion directories and must not modify `.bashrc`, `.zshrc`, or Fish configuration files automatically. If additional shell configuration is required, the installer should print an actionable message.

Shell startup files are user-managed configuration. Automatically rewriting them introduces behavior users may not expect and requires the installer to handle marker parsing, upgrades, locking, atomic replacement, symlinks, file permissions, and uninstall restoration. Most of the complexity added by this PR exists solely because it chose to edit `.zshrc`.

Please install the Zsh completion file into the standard user completion directory instead, for example:

```text
~/.local/share/zsh/site-functions/_aenv
```

If that directory is not present in the user’s `fpath`, the installer should print an actionable message such as:

```zsh
fpath=(~/.local/share/zsh/site-functions $fpath)
autoload -Uz compinit
compinit
```

The user can then decide whether and where to add that configuration. Installation and removal should only manage the completion file itself, not the user’s shell startup files.

## 2. The installer implementation is disproportionately large and duplicated

This PR is described as a static completion scaffolding change, but it currently adds approximately 2,300 lines. A major part of that comes from copying the same roughly 500-line `aenv_completion_install` implementation into three files:

- `scripts/shell-completion.sh`
- `scripts/install-cli.sh`
- `scripts/install.sh`

That is approximately 1,500 lines of duplicated shell code. The duplication then requires an additional synchronization checker, dedicated filesystem-management tests, and a separate CI job.

This is disproportionate to the feature being implemented. The required behavior is relatively small:

- install three lightweight completion loaders into standard directories;
- remove the files during uninstall;
- avoid overwriting unrelated files;
- cover basic installation, upgrade, and removal behavior;
- print configuration guidance when necessary.

Without editing `.zshrc` or implementing a general-purpose shell configuration file manager, I would expect the automatic installation portion to be roughly **200–300 lines including its tests**. This is an estimate rather than a strict line-count requirement, but if a robust implementation cannot reasonably stay within approximately **300–400 lines**, that is a strong indication that it should be reviewed separately.

Therefore, either of the following approaches would be acceptable:

### Option A: Keep automatic installation in this PR

This is fine if the implementation can be significantly simplified:

- do not modify shell startup files;
- install only small files into the standard completion directories;
- remove the large duplicated helper and its synchronization machinery;
- keep the installation implementation and tests within a reasonably small scope.

### Option B: Split the work into two PRs

If automatic installation still requires substantial installer-specific work, please split it:

1. This PR contains:
   - `aenv completion bash|zsh|fish`;
   - visible aliases;
   - Rust tests;
   - manual activation documentation.

2. A follow-up PR contains:
   - automatic installation and removal;
   - standard completion-directory integration;
   - installer-specific tests and documentation.

Again, I appreciate that the combined scope came from my earlier request. The concern is not that automatic installation was included, but that the current solution has grown into a large and duplicated shell file-management subsystem. Please either simplify that portion substantially or move it into a focused follow-up PR.

rbalachandar:
> ### Option B: Split the work into two PRs
> If automatic installation still requires substantial installer-specific work, please split it:
> 
> 1. This PR contains:
>    
>    * `aenv completion bash|zsh|fish`;
>    * visible aliases;
>    * Rust tests;
>    * manual activation documentation.
> 2. A follow-up PR contains:
>    
>    * automatic installation and removal;
>    * standard completion-directory integration;
>    * installer-specific tests and documentation.
> 
> Again, I appreciate that the combined scope came from my earlier request. The concern is not that automatic installation was included, but that the current solution has grown into a large and duplicated shell file-management subsystem. Please either simplify that portion substantially or move it into a focused follow-up PR.

Removed installation related changes for separate PR.
Draft: https://github.com/kvcache-ai/AgentENV/pull/115

rbalachandar:
@yingdi-shan adding a comment/reason for closing would be helpful, so it is clear that the pr no longer relevant.

LSX-s-Software:
@rbalachandar Sorry, we mistakenly thought this PR had been replaced by #115. We've reopened this PR and we will review it soon. Thanks for contributing.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,52 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
+aenv completion fish
+```
+
+Each command writes the matching script to stdout, so redirect it into the standard per-user completion directory for your shell:
+
+```bash
+aenv completion bash > ~/.local/share/bash-completion/completions/aenv
+aenv completion zsh  > ~/.local/share/zsh/site-functions/_aenv
+aenv completion fish > ~/.config/fish/completions/aenv.fish
+```
```
LSX-s-Software: The redirect examples assume that all parent directories already exist. On a fresh environment, commands such as:

```bash
aenv completion bash > ~/.local/share/bash-completion/completions/aenv
```

fail with `No such file or directory`.

docs/src/getting-started/aenv-cli.md
```diff
@@ -274,3 +274,52 @@ aenv snapshot list --sandbox-id <sandbox-id>
...
+
+bash and fish auto-load completions from those directories. zsh autoloads `_cmdname` functions from directories on `fpath`; `~/.local/share/zsh/site-functions` is not on `fpath` by default, so if completion does not load, add the directory to `fpath` before `compinit`:
```
LSX-s-Software: Also, “bash auto-loads completions from this directory” is slightly misleading: this behavior is provided by `bash-completion`, not Bash itself, and requires `bash-completion` to be installed and initialized. Please clarify that dependency in the documentation.

LSX-s-Software (changes requested):
Thanks for correcting the completion paths. Several documentation issue remains. You can use the following markdown to replace the sections starting from _Generate a script_:

````markdown
### Generate and install a script

#### Bash

```bash
mkdir -p ~/.local/share/bash-completion/completions
aenv completion bash > ~/.local/share/bash-completion/completions/aenv
```

The Bash completion file is loaded on demand by
[`bash-completion`](https://github.com/scop/bash-completion).
This requires `bash-completion` to be installed and initialized in the current
shell.

#### Zsh

```zsh
mkdir -p ~/.local/share/zsh/site-functions
aenv completion zsh > ~/.local/share/zsh/site-functions/_aenv
```

Zsh loads completion functions from directories listed in `fpath`.
`~/.local/share/zsh/site-functions` is not included in `fpath` by default on
all systems. Add the following lines to `~/.zshrc` before any existing
`compinit` invocation:

```zsh
fpath=(~/.local/share/zsh/site-functions $fpath)
autoload -Uz compinit
compinit
```

#### Fish

```shell
mkdir -p ~/.config/fish/completions
aenv completion fish > ~/.config/fish/completions/aenv.fish
```

### Activate without installing

To test completion for the current shell session without saving a generated
file:

```bash
source <(aenv completion bash)         # Bash
eval "$(aenv completion zsh)"          # Zsh
aenv completion fish | source          # Fish
```

Once loaded, completion covers the CLI surface:

```bash
aenv <TAB>                       # top-level commands
aenv snapshot <TAB>              # nested subcommands
aenv list --output <TAB>         # enum values: table, json
aenv build ./<TAB>               # local path arguments
```
````

docs/src/getting-started/aenv-cli.md
rbalachandar: Addressed in commit a13c510: each Bash, Zsh, and Fish example now creates its parent directory with mkdir -p before redirecting the generated completion file.

docs/src/getting-started/aenv-cli.md
rbalachandar: Addressed in commit a13c510: the Bash section now explains that loading is provided by bash-completion and requires bash-completion to be installed and initialized.

LSX-s-Software (approved):

PR #91, opened 2026-07-31
docs: add static multi-node deployment guide

### LINKED ISSUE #59: Multi-node deployment without Kubernetes
Hi, I'm planning to deploy AgentENV across ~10 physical machines for agentic RL training.

The documentation covers two multi-node paths:
- **Docker Compose** — described as "Multi-Node Simulation" (single host only)
- **Kubernetes / k3s** — the only documented true multi-machine option

My question: is there a supported way to run a real multi-node cluster **without Kubernetes**?

Specifically, I'm wondering if the following approach would work:
- Run `aenv-node` via the install script + systemd on each physical machine
- Run Gateway and Scheduler as standalone processes (or Docker containers) on a control machine
- Configure each node to report heartbeats to the Scheduler via `AENV_OBSERVABILITY_SCHEDULER_ENDPOINT`

Or is Kubernetes/k3s currently the only supported path for multi-machine deployments?

Thanks!

catyans:
## Summary

- document a real multi-machine AgentENV deployment using static Scheduler discovery
- provide a complete control-plane JSON example, runtime heartbeat wiring, and systemd units
- document network flows, verification, operations, and common identity/addressing failures
- distinguish this deployment from the single-host Docker Compose simulation

## Why

Docker Compose only simulates multiple nodes on one host, while the existing
Kubernetes guide is the only documented multi-machine path. AgentENV already
supports a non-Kubernetes deployment with statically configured runtime nodes,
but the required identity, endpoint, and heartbeat relationship was not
documented.

The guide makes the current constraint explicit: heartbeats do not
auto-register unknown nodes, so every `AENV_NODE_ID` must match a
`scheduler.nodes[].id`, and static membership changes require a Scheduler
restart.

## Validation

- `git diff --check`
- `make -C services test` on the Lingjun test host
- built both Go control-plane binaries on the Lingjun test host
- extracted the JSON example from the guide and started Scheduler and Gateway
  with isolated ports
- verified the Gateway `/health` and `/nodes` endpoints through the temporary
  control plane

Closes #59.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -11,6 +11,7 @@
 
 - [Docker (Single Node)](./deployment/docker.md)
 - [Docker Compose (Multi-Node Simulation)](./deployment/docker-compose.md)
+- [Static Multi-Node (Without Kubernetes)](./deployment/static-multi-node.md)
 - [Kubernetes (Multi-Node)](./deployment/kubernetes.md)
 - [Manual Compile (Single Node)](./deployment/manual-compile.md)
 
--- a/docs/src/deployment/docker-compose.md
+++ b/docs/src/deployment/docker-compose.md
@@ -2,6 +2,9 @@
 
 Run a full multi-node stack on a single host using Docker Compose. This simulates a production-like topology with a gateway, scheduler, and multiple AgentENV backend nodes.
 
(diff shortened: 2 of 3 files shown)


yingdi-shan (approved):
LGTM. Thanks for your contribution.

Issue #101, opened 2026-07-31
[RFC] Exploring an end-to-end ARM64 support path for AgentENV: findings from a validated DGX Spark deployment


leether:
### Affected component

Cross-cutting / other

### RFC intent and possible closure

This is a request for design feedback, not a request to declare ARM64 supported
or to merge one large implementation. A useful outcome could be any of the
following:

1. ARM64 server/runtime support is currently out of scope; the issue can close
   with that position recorded.
2. The project is open to an initial support tier; the issue can close after
   that tier and its compatibility boundary are recorded and focused follow-up
   issues or PRs are identified.
3. The project needs more evidence before choosing a direction; the issue can
   close or be deferred with the missing validation and revisit conditions
   recorded.

### Problem statement

This RFC asks whether ARM64 server/runtime support is within AgentENV's scope
and, if so, what initial level of support the project might intend to provide
and test.

A server that compiles on `aarch64` is an important milestone, but it does not
by itself prove that the selected Firecracker build, guest kernel, tools drive,
ublk/OverlayBD path, seccomp policy, and snapshot lifecycle work together.

There are two adjacent open contributions as of 2026-07-31:

- [#86](https://github.com/kvcache-ai/AgentENV/pull/86) proposes ARM64 release
  artifacts, installer selection, and multi-architecture container publishing.
- [#94](https://github.com/kvcache-ai/AgentENV/pull/94) proposes an ARM64
  source-build path, architecture-aware Firecracker and guest-kernel downloads,
  ARM tools-drive construction, and openEuler recognition.

This issue is not a replacement for either PR. It first asks whether the
project would like to pursue ARM64 server/runtime support. If the answer is
yes, this case study may help distinguish a build or installation milestone
from an end-to-end supported runtime.

We independently assembled and operated one complete AgentENV stack on an
NVIDIA DGX Spark. This work did not use or validate #86 or #94. It demonstrates
that one full ARM64 combination is feasible, but it does not prove that every
ARM64 host, kernel, or distribution is compatible.

### Related work and scope

| Item | What it covers | What remains outside its stated validation |
| --- | --- | --- |
| [#60](https://github.com/kvcache-ai/AgentENV/issues/60) | General request for ARM support | The intended support boundary and acceptance criteria |
| [#63](https://github.com/kvcache-ai/AgentENV/issues/63) | Template build failure caused by using an incompatible upstream Firecracker binary | A complete compatible dependency set and snapshot regression |
[...]

PR #103, opened 2026-08-01
fix(sandbox): isolate envd bootstrap connections across generations

### LINKED ISSUE #102: Snapshot restore may stall when AgentENV reuses an envd bootstrap connection across sandbox generations
## Affected component

Orchestrator / sandbox lifecycle

## AgentENV version

- Version or commit: reproduced on [`leether/AgentENV@105fbbc`](https://github.com/leether/AgentENV/commit/105fbbc62cec0f00291de30a62038968a0e4dd11), based on [`kvcache-ai/AgentENV@8174bd9`](https://github.com/kvcache-ai/AgentENV/commit/8174bd9674806917a2c0e75d26a30b409459c51d)
- Local modifications: ARM64 assembly-line changes were required to run AgentENV on this host. Before the proposed fix, `src/sandbox/envd.rs` was unchanged from upstream at the affected shared-client code path.
- First source revision known to contain the affected shared-client pattern: [`kvcache-ai/AgentENV@8f028b1`](https://github.com/kvcache-ai/AgentENV/commit/8f028b1369a4ce9a66d140174d0a646e72398517) (initial open-source release; not separately runtime-tested for this report)
- envd: `0.6.13`, built from source tree [`e2b-dev/infra@bfdbb24`](https://github.com/e2b-dev/infra/commit/bfdbb24130ede13b449331acab40d39981f9b329)

## Host environment

- OS: Ubuntu 24.04.4 LTS
- Kernel (`uname -a`, hostname redacted): `Linux <dgx-host> 6.17.0-1021-nvidia #21-Ubuntu SMP PREEMPT_DYNAMIC Wed May 27 19:14:05 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux`
- Architecture: `aarch64`, 4 KiB pages
- Bare metal, VM, or cloud: NVIDIA DGX Spark, bare metal
- ublk support: `/dev/ublk-control` present; real `uvm-ublk-daemon` and OverlayBD paths used
- CPU: 20 CPUs, Cortex-X925 / Cortex-A725
- Memory: 121 GiB visible to the OS
- Storage/filesystem: NVMe, root filesystem on ext4 (`/dev/nvme0n1p2`)
- Firecracker: `kvcache-ai/firecracker-next` `v1.17.0-next.1` lineage, exact downstream build `b792de1c4618092b0d712c59e5d2a195c97ff5bd`

## Operation and configuration

The failure was found by running the repository snapshot benchmark repeatedly in fresh processes. The benchmark uses a 1-vCPU, 128-MiB guest and calls `FirecrackerSandbox::resume_from_snapshot_config()` directly.

Representative invocation shape, with isolated paths and test CIDRs:

```shell
# Build the repository benchmark first.
cargo bench --bench snapshot --no-run

# Run the built snapshot benchmark with an isolated AgentENV environment.
# The exact target hash depends on the build.
env \
  AENV_CONFIG_PATH=/tmp/agentenv-official-bench.toml \
  AENV_HOME_PATH=/tmp/agentenv-official-bench/home \
  AENV_RUNTIME_PATH=/tmp/agentenv-official-bench-run1/runtime \
  AENV_DEPS_PATH=/path/to/aenv-deps \
  AENV_FIRECRACKER_WORK_DIR=/tmp/agentenv-official-bench-run1/work \
  AENV_FIRECRACKER_SERIAL_DIR=/tmp/agentenv-official-bench-run1/serial \
  NO_PROXY=127.0.0.1,localhost,10.211.0.0/16,10.212.0.0/16,169.254.0.0/16 \
  no_proxy=127.0.0.1,localhost,10.211.0.0/16,10.212.0.0/16,169.254.0.0/16 \
  ./target/release/deps/snapshot-<build-hash> snapshot_resume
```

Relevant configuration characteristics:

```toml
[pool]
low_watermark = 2
high_watermark = 64

[pool.network]
maintenance_enabled = true
```

The test used dedicated home/runtime/work/serial/cache directories, an isolated ublk daemon and metrics port, and dedicated network CIDRs. Production AgentENV was not stopped or restarted.

## Steps to reproduce

1. Run an AgentENV server process containing the process-wide `ENVD_HTTP_CLIENT` from `src/sandbox/envd.rs`.
2. Create a Firecracker snapshot after envd bootstrap has created guest TCP state.
3. Repeatedly restore the same snapshot in the same AgentENV process, allowing warm network slots and interaction IPs to be reused.
4. Ensure the restore sequence returns to the interaction IP/slot whose TCP tuple is present in the snapshot. The repository `snapshot` benchmark reaches this state through normal slot rotation; a one-slot/deterministic test setup can force it directly.
5. Observe one envd health/init request reuse the host-side idle connection from the previous runtime generation.

In three fresh benchmark processes, each cold and hot resume group contained one approximately 30-second sample. A separate deterministic packet-capture run reproduced the same generation sequence with timings of `163.237 ms`, `30,995.648 ms`, and `160.390 ms`.

## Expected behavior

Each sandbox runtime generation should bootstrap envd through a TCP connection that belongs to that generation. Restoring a snapshot into a reused interaction IP should not allow the host HTTP client to send health or init over an idle connection retained from a previous VM.

If an old flow is unusable, bootstrap should establish a fresh connection immediately rather than waiting for the TCP user timeout.

## Actual behavior

The process-wide reqwest client pools by HTTP origin (`scheme + IP + port`) and outlives individual sandbox runtimes. When the same interaction IP is reused, it can check out an idle TCP connection from the previous runtime generation.

If the restored guest snapshot contains the matching TCP tuple, the guest does not immediately reject the packet. Instead, guest and host are in different TCP sequence spaces and make no forward progress. The bootstrap request waits approximately 30 seconds before reqwest opens a new connection.

Across three unforced benchmark rounds:

- cold snapshot resume: 3/30 samples stalled for `30.91–31.04 s`; the other 27 were `91.70–118.71 ms`;
- hot snapshot resume: 3/30 samples stalled for `30.52–30.54 s`; the other 27 were `61.55–85.06 ms`.

An ordinary-template control ran 10 starts in `95–179 ms`. In that case the immutable snapshot did not contain the pooled connection's tuple, so the guest returned RST and reqwest reconnected quickly. Reusing the IP alone is therefore not sufficient; the snapshot must also contain the matching guest-side TCP state.

## Logs and diagnostics

The deterministic trace used this tuple:

```text
10.242.0.8:39158 -> 10.241.0.4:49983
```

Before the snapshot, the connection was at guest/host state `G0/H0`. The first restore reused the tuple successfully and advanced the host side to `H1`:

```text
host health  seq 42026036:42026097 ack 1773910168
guest reply  seq 1773910168:1773910287 ack 42026097
host init    seq 42026097:42026255 ack 1773910287
guest reply  seq 1773910287:1773910414 ack 42026255
```

The second restore rewound only the guest to `G0`, while the pooled host socket remained at `H1`:

```text
host:  seq 42026255:42026316 ack 1773910414
guest: seq 1773910041:1773910168 ack 42026036
```

The host sequence is ahead by exactly 219 bytes, matching the first restore's 61-byte health request plus 158-byte init request. Both sides continued ACK/SACK and retransmission without advancing the cumulative ACK.

The first stuck packet was captured at `13:32:46.595216`; the fresh-connection SYN was sent at `13:33:17.457549`, a `30.862333 s` interval. The built reqwest `0.13.3` has a 30-second TCP user timeout. As a control, changing only that timeout to five seconds shortened the stall to approximately 5.5 seconds.

Further controls:

- disabling idle pooling removed the stall;
- constructing a client per runtime generation also removed the stall;
- PAWS/timestamp rejection was excluded because both sides continued exchanging ACK/SACK and timestamp values advanced;
- killing the previously observed unrelated ghost TCP flows did not remove the earlier 504 incident, so this report does not claim that every restored stale socket causes this bootstrap stall.

## Reproduction frequency

Intermittently

With the unforced repository benchmark and normal slot rotation, it appeared in 3/30 cold and 3/30 hot samples. Once the same-slot generation sequence was made deterministic, that sequence reproduced it every time.

## Regression information

- Last known good: unknown
- First known bad: unknown as a runtime regression
- First source revision known to contain the affected process-wide client: [`kvcache-ai/AgentENV@8f028b1`](https://github.com/kvcache-ai/AgentENV/commit/8f028b1369a4ce9a66d140174d0a646e72398517) (not separately runtime-tested)

### Root cause

The connection identity used by the host pool is the HTTP origin, but the actual endpoint identity is `(origin, sandbox runtime generation)`. Snapshot restore rewinds guest TCP state; the host process and its idle connection pool are not part of the snapshot and continue forward. Warm network-slot reuse then assigns the same origin to a later guest runtime.

The resulting state transition is:

```text
snapshot:       guest G0 / host H0
first restore:  guest G1 / host H1
second restore: guest G0 / host H1  -> sequence mismatch -> TCP user timeout
```

PR #56 addressed the same origin-versus-runtime-generation mismatch for the sandbox HTTP proxy client. The envd bootstrap path uses a separate generated reqwest client and was not covered by that change.

This was discovered while following up the unexplained approximately 30-second snapshot benchmark samples reported in #101. ARM64 is the environment in which it was found, but the mechanism is not ARM-specific: it requires snapshot restore, a host client that outlives the guest runtime, and reuse of the same interaction address.

### Narrow fix evaluated

We tested keeping the shared bootstrap client configuration while preventing its connections from entering the idle pool:

```rust
Client::builder()
    .pool_max_idle_per_host(0)
    .build()
```

The scope is only envd `GET /health` and `POST /init`. Command execution continues to use a separately created gRPC `ProcessClient`; sandbox proxy traffic is unchanged.

A behavioral regression test keeps the health socket open and requires init to arrive through a second `accept()`. With `pool_max_idle_per_host(1)` the negative control times out because init reuses the first socket; with `0` it passes.

Validation of the narrow fix on the same isolated ARM64 host:

- three rounds of 10 cold + 10 hot samples: 60/60 below 200 ms;
- 1,000 sequential hot restores: p50 `130.00 ms`, p99 `152.91 ms`, max `158.41 ms`, `0/1,000 > 1 s`;
- two 50-concurrency x 10 runs: 1,000/1,000 restores completed;
- library tests: 660 passed, 0 failed, 2 ignored;
- short-lived sockets returned to zero orphan count within 20 seconds after the concurrent run; allocated file handles returned close to baseline.

The evaluated patch intentionally does not add `.no_proxy()` or `.http1_only()`, because proxy routing and protocol negotiation are unrelated behavior changes.

Would maintainers be open to a focused PR that applies `pool_max_idle_per_host(0)` only to the envd bootstrap client and adds the behavioral regression test? I can prepare it on a clean branch from current upstream `main`.

## Pre-submission checklist

- [x] I searched existing open and closed issues and did not find a duplicate.
- [x] I provided a minimal reproduction or explained why one cannot be provided.
- [x] I removed credentials, tokens, private image references, and other sensitive information.
- [x] This is not a security vulnerability.

leether:
## What

- disable idle TCP retention for the shared envd bootstrap HTTP client;
- keep the change limited to envd `GET /health` and `POST /init`;
- add a bounded behavioral regression test that requires the two bootstrap requests to use different TCP connections.

## Why

AgentENV reuses interaction IPs across sandbox runtime generations, while the process-wide reqwest client pooled idle connections by HTTP origin (`scheme + IP + port`). The pool key does not include the sandbox runtime generation.

After snapshot restore, a later runtime can therefore receive the same interaction IP while the host client still holds a TCP connection from the previous runtime. If the restored guest also contains the matching TCP tuple, the host and guest can be in different TCP sequence spaces. An envd health or init request may then stall until the TCP user timeout instead of establishing a fresh connection.

The packet-level reproduction, controls, and full stress evidence are in #102. The issue was discovered during ARM64 snapshot validation in #101, but the lifecycle condition is not architecture-specific.

## Related issue

Closes #102

Related:

- #56 fixed the same origin-versus-runtime-generation mismatch for the separate sandbox HTTP proxy client;
- #101 is the ARM64 assembly-line RFC where this bootstrap stall was discovered.

## Scope and non-goals

This PR changes only `src/sandbox/envd.rs`:

- the envd bootstrap HTTP client no longer retains idle connections;
- the client object remains shared, so construction and configuration are not repeated per sandbox;
- the regression test covers observable connection behavior without inspecting reqwest internals.

This PR does not change:

- sandbox HTTP proxy traffic;
- gRPC process execution;
- Firecracker, the guest kernel, envd, or `tools.ext4`;
- snapshot formats, artifact layouts, or network-slot allocation;
- proxy environment handling or HTTP protocol selection.

The shared bootstrap client is now built with:

```rust
Client::builder()
    .pool_max_idle_per_host(0)
    .build()
```

For a successful bootstrap, health and init therefore use separate TCP connections. Failed health retries also open fresh connections. This prevents a bootstrap request from checking out a socket retained from a previous VM generation.

Command execution remains on the separately created gRPC `ProcessClient`, so repeated exec calls are unaffected. The patch intentionally does not add `.no_proxy()` or `.http1_only()` because proxy routing and protocol negotiation are independent behaviors.

The regression test:

1. answers health over a keep-alive socket and holds that socket open as the previous-generation connection;
2. treats a second `accept()` as connecting to the next VM assigned the same address;
3. requires init to arrive on the second connection;
4. bounds individual steps and the complete interaction with timeouts.

With `pool_max_idle_per_host(1)` as a local negative control, the test fails because init reuses the first socket. With `0`, it passes.

- Public API or generated protocol: N/A — no API or generated-code changes.
- Configuration or defaults: N/A — no new configuration or changed default.
- Snapshot manifest, artifact layout, or storage format: N/A — no persisted-format changes.
- Upgrade and rollback: no migration is required; deploying the updated server enables the behavior, and rollback is the previous server binary or a revert of this commit.
- Host requirements, permissions, ports, or dependencies: N/A — unchanged.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [ ] Documentation updated

Commands and results:

```text
make fmt
  passed

make clippy
  passed for the full workspace with all targets and all features

make AENV_TEST_STATE_ID=pr102-arm64-20260801 \
  "CAPABILITY_RUNNER=CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER=$PWD/scripts/run-with-capabilities.sh" \
  test-unit
  passed: agentenv 671, linux-cap 3, privileged agentenv 4,
  uvm-ublk 10, uvm-ublk-daemon 46, plus both verification scripts

cargo test -p agentenv --lib sandbox::envd::tests
  passed: 2

local negative control: pool_max_idle_per_host(1)
  failed in 2.03 s with the expected
  "init reused the health connection instead of opening a new one"

official snapshot benchmark, three isolated rounds of 10 cold + 10 hot
  passed: 60/60 completed; no sample exceeded 1 s; no approximately 30 s stall
```

The benchmark is evidence that the observed stall is removed, not a controlled A/B attribution of sub-second latency. Post-fix cold means were `168.50–182.34 ms`; hot means were `119.81–124.81 ms`. Full sample ranges and the deeper sequential/concurrent stress results are recorded in #102.

Skipped checks and reasons:

- Separate Rust integration targets were not rerun on this clean commit; the focused behavioral test and the real snapshot benchmark exercise the changed lifecycle path. The same change had previously passed `fc::multiple_resumes_have_independent_disk_state` downstream.
- `make -C services test` was not run because `services/` is unchanged.
- Code generation was not run because no schema or generated file changed.
- Documentation was not changed because the user-facing API and configuration are unchanged.
- On ARM64, the current Makefile's default capability runner variable is x86_64-specific. `make test-unit` was therefore run with the command-line `CAPABILITY_RUNNER` override shown above; this PR intentionally does not include an unrelated Makefile change.

## Risks and reviewer notes

- Bootstrap pays one additional host-to-guest TCP handshake, and failed health retries also use fresh sockets. The affected client performs only health/init; sustained exec and proxy traffic use other clients.
- Short-lived bootstrap sockets can enter `TIME_WAIT`. The concurrent stress run reported in #102 returned to zero orphan sockets within 20 seconds and file-handle allocation returned close to baseline.
- A per-runtime client also prevents the failure, but it would require threading runtime identity through this generated-client path solely to preserve one health-to-init idle connection. Disabling idle retention is narrower for a two-request bootstrap path.
- The complete mock interaction has a five-second outer timeout, so connector or handshake regressions fail CI instead of hanging it.
- Reviewer focus: the client construction and `bootstrap_http_client_does_not_reuse_connections` test in `src/sandbox/envd.rs`.

## Checklist

### DIFF
--- a/src/sandbox/envd.rs
+++ b/src/sandbox/envd.rs
@@ -13,7 +13,14 @@ use envd::reqwest::Client;
 
 const HEALTH_PROBE_TIMEOUT: Duration = Duration::from_secs(1);
 
-static ENVD_HTTP_CLIENT: LazyLock<Client> = LazyLock::new(Client::new);
+// Bootstrap addresses can be reused across sandbox runtime generations. Do not
+// retain connections that may belong to the previous VM assigned the same IP.
+static ENVD_BOOTSTRAP_HTTP_CLIENT: LazyLock<Client> = LazyLock::new(|| {
+    Client::builder()
+        .pool_max_idle_per_host(0)
+        .build()
+        .expect("build envd bootstrap HTTP client")
+});
 
 pub(crate) struct EnvdInstance {
     config: Configuration,
@@ -24,11 +31,12 @@ impl EnvdInstance {
     pub(crate) fn new(base_path: String) -> Self {
         let grpc_address = base_path.clone();
         Self {
-            // Use full construction here to ensure the shared `Client` is used and no new instances are created.
+            // Share client configuration without retaining bootstrap TCP
+            // connections across sandbox runtime generations.
             config: Configuration {
                 base_path,
                 user_agent: None,
-                client: ENVD_HTTP_CLIENT.clone(),
+                client: ENVD_BOOTSTRAP_HTTP_CLIENT.clone(),
                 basic_auth: None,
                 oauth_access_token: None,
                 bearer_access_token: None,
@@ -130,9 +138,12 @@ impl EnvdInstance {
 
 #[cfg(test)]
 mod tests {
+    use std::net::SocketAddr;
     use std::time::Instant;
 
-    use tokio::net::TcpListener;
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
+    use tokio::net::{TcpListener, TcpStream};
+    use tokio::time::timeout;
 
     use super::*;
 
@@ -158,6 +169,113 @@ mod tests {
         server.abort();
         assert!(error.to_string().contains("timed out waiting for envd"));
         assert!(started.elapsed() < Duration::from_millis(500));
+        Ok(())
+    }
+
+    async fn read_request(socket: &mut TcpStream) -> Result<String> {
+        let mut request = Vec::with_capacity(1024);
+        let mut chunk = [0_u8; 1024];
+
+        let (header_end, content_length) = loop {
+            let bytes_read = socket.read(&mut chunk).await?;
+            if bytes_read == 0 {
+                return Err(anyhow!("connection closed before request headers"));
+            }
+            request.extend_from_slice(&chunk[..bytes_read]);
+
+            if request.len() > 64 * 1024 {
+                return Err(anyhow!("request headers exceed test limit"));
+            }
+
+            let Some(header_end) = request.windows(4).position(|window| window == b"\r\n\r\n")
+            else {
+                continue;
+            };
+            let header_text = std::str::from_utf8(&request[..header_end + 4])?;
+            let content_length = header_text
+                .lines()
+                .find_map(|line| {
+                    let (name, value) = line.split_once(':')?;
+                    name.eq_ignore_ascii_case("content-length")
+                        .then(|| value.trim().parse::<usize>().ok())
+                        .flatten()
+                })
+                .unwrap_or(0);
+            break (header_end + 4, content_length);
+        };
+
+        while request.len() < header_end + content_length {
+            let bytes_read = socket.read(&mut chunk).await?;
+            if bytes_read == 0 {
+                return Err(anyhow!("connection closed before request body"));
+            }
+            request.extend_from_slice(&chunk[..bytes_read]);
+        }
+
+        let request_line = std::str::from_utf8(&request)?
+            .lines()
+            .next()
+            .ok_or_else(|| anyhow!("request line missing"))?;
+        Ok(request_line.to_owned())
+    }
+
+    // The first accepted socket models an idle connection retained from the
+    // previous runtime generation. A second accept models connecting to the
+    // next VM assigned the same address; bytes sent on the first socket would
+    // therefore be stale cross-generation reuse.
+    async fn serve_two_bootstrap_requests(listener: TcpListener) -> Result<Vec<SocketAddr>> {
+        const RESPONSE: &[u8] =
+            b"HTTP/1.1 204 No Content\r\nContent-Length: 0\r\nConnection: keep-alive\r\n\r\n";
+        let mut peer_addresses = Vec::with_capacity(2);
+        let mut held_connections = Vec::with_capacity(2);
+
+        for expected_prefix in ["GET /health ", "POST /init "] {
+            let (mut socket, peer_address) = listener.accept().await?;
+            let request_line = read_request(&mut socket).await?;
+            if !request_line.starts_with(expected_prefix) {
+                return Err(anyhow!(
+                    "expected request prefix {expected_prefix:?}, got {request_line:?}"
+                ));
+            }
+            socket.write_all(RESPONSE).await?;
+            peer_addresses.push(peer_address);
+
+            // Keep the first socket open so a pooling client would send init
+            // there and never reach the second accept.
+            held_connections.push(socket);
+        }
+
+        Ok(peer_addresses)
+    }
+
+    async fn run_bootstrap_no_reuse_interaction() -> Result<()> {
+        let listener = TcpListener::bind("127.0.0.1:0").await?;
+        let address = listener.local_addr()?;
+        let server = tokio::spawn(serve_two_bootstrap_requests(listener));
+        let envd = EnvdInstance::new(format!("http://{address}"));
+
+        envd.wait_for_ready(Duration::from_secs(2), Duration::from_millis(10))
+            .await?;
+        timeout(Duration::from_secs(2), envd.init(None, None, None))
+            .await
+            .map_err(|_| {
+                anyhow!("init reused the health connection instead of opening a new one")
+            })??;
+
+        let peer_addresses = timeout(Duration::from_secs(2), server)
+            .await
+            .map_err(|_| anyhow!("timed out waiting for two bootstrap connections"))?
+            .map_err(|err| anyhow!("mock envd task failed: {err}"))??;
+        assert_eq!(peer_addresses.len(), 2);
+        assert_ne!(peer_addresses[0].port(), peer_addresses[1].port());
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn bootstrap_http_client_does_not_reuse_connections() -> Result<()> {
+        timeout(Duration::from_secs(5), run_bootstrap_no_reuse_interaction())
+            .await
+            .map_err(|_| anyhow!("timed out running complete envd bootstrap interaction"))??;
         Ok(())
     }
 }


yingdi-shan (approved):
LGTM. Thanks for your contribution!

PR #108, opened 2026-08-03
Fix sandbox lifecycle bugs and improve network policy handling

### LINKED ISSUE #42: Direct Firecracker start failures can retain sandbox resources

### LINKED ISSUE #99: Pause allocation failures leave the sandbox detached in Pausing
## What happened?

If paused-sandbox artifact-root allocation fails, `pause_sandbox` returns an error after it has already removed the sandbox's runtime handle and proxy route. The metadata remains in `Pausing`, so the still-running sandbox is no longer reachable and a later pause cannot retry the operation.

This was reproduced at `f41abb21324f6b0520abf34b7720aa260ddd10eb` with the in-memory orchestrator and a persister that deterministically fails its next `allocate_artifact_root` call. No KVM or privileged host setup is required.

## Where does it happen?

`pause_sandbox_inner` protects the paused runtime artifacts and then calls `detach_sandbox_handle_and_route` before allocating the artifact root:

- `src/orchestrator/service.rs`: the handle and route are detached immediately before `SandboxPersister::allocate_artifact_root`.
- On the baseline, allocation uses `await?`, so its error bypasses the rollback implemented for later `sandbox.pause(...)` and persistence failures.

The resulting control flow is:

```text
Running
   |
   v
Pausing -- protect PausedSandbox refs
   |
   +-- detach runtime handle
   +-- remove proxy route
   |
   v
allocate_artifact_root() fails
   |
   +-- returns error without rollback
   v
metadata: Pausing   runtime: still alive but detached   route: absent
```

## How can we reproduce it?

1. Construct an orchestrator with `RecordingPersister`.
2. Configure `RecordingPersister::fail_next(RecordingCall::AllocateArtifactRoot)`.
3. Create a sandbox and call `pause_sandbox`.
4. Observe that the call returns `SandboxPersistenceFailed`.
5. On the baseline, inspect the postcondition: the metadata is not restored to `Running`, the runtime handle and proxy route have already been removed, and the temporary `PausedSandbox` image ownership is not released by this path.

A regression test can exercise the behavior entirely through the existing mock backend and persister seam. The important assertion is not only that the first call fails: after rollback, the sandbox must be routable and a second `pause_sandbox` call must succeed.

## What did you expect to happen?

Artifact-root allocation happens before the backend mutates the running VM, so this failure should be recoverable. The operation should:

1. restore the runtime handle;
2. restore the removed proxy route;
3. transition metadata from `Pausing` back to `Running`;
4. release the temporary `PausedSandbox` image references; and
5. return the allocation error while leaving a subsequent pause retryable.

If restoring metadata itself fails, the returned error should preserve both the allocation failure and rollback failure rather than reporting a clean recovery.

## Acceptance criteria

- A deterministic unit test forces artifact-root allocation to fail after handle/route detachment.
- After the failure, metadata is `Running` with no paused state, the runtime handle is present, and the proxy route is ready.
- The temporary `RuntimeImageOwner::PausedSandbox` reference is released.
- No paused state is persisted for the failed attempt.
- Retrying pause succeeds and reaches `Paused`.
- Existing rollback behavior for backend pause failures and persistence failures is unchanged.

## Anything else we need to know?

This report is scoped to the recoverable allocation failure before `SandboxBackend::pause` is invoked. It does not propose changing terminal backend pause failures, stop-failure handling, or restart-time persistence cleanup.

Environment for the deterministic reproduction: Linux x86_64, Rust stable toolchain. The proof is unit-level and does not require `/dev/kvm` because the failure occurs before backend pause execution.

LSX-s-Software:
## What

Fixes three sandbox lifecycle bugs (network-slot leak on egress-policy failure, unhandled artifact-root allocation error during pause, and cancellation-unsafe network/custom-extension updates) and makes egress policy replacement atomic via a single `iptables-restore` transaction.

## Why

Several lifecycle paths left the sandbox in an inconsistent state on failure or cancellation:
- A failed `set_egress_policy` call dropped the network slot before it was attached to `self.network_slot`, leaking the underlying network resources.
- `allocate_artifact_root` failure during pause occurred after the handle/route were already detached, with no rollback of image refs or sandbox state.
- `replace_sandbox_network_policy` and `patch_sandbox_custom_extension_params` were not wrapped in the cancellation-safe execution path used by other lifecycle operations (create/fork/delete/pause/resume/snapshot), so a client disconnecting mid-request could leave the update half-applied.
- Egress policy updates flushed the `USER_EGRESS_CHAIN` and installed new rules as two separate `iptables-restore` invocations, leaving a brief window with inconsistent (or no) egress rules.

## Related issue

Closes #42
Closes #99

## Scope and non-goals

Included: sandbox network-slot leak fix, pause artifact-root allocation error handling, cancellation-safety for network policy/custom extension param updates, atomic egress policy replacement.
Excluded: no serialization for concurrent network policy/custom extension param updates.

- `FirecrackerSandbox`: `network_slot` is now assigned to `self.network_slot` **before** `set_egress_policy` is invoked, so a failed call still leaves the slot reachable by the normal cleanup path instead of being dropped in place.
- `OrchestratorService::pause_sandbox_inner`: `allocate_artifact_root` is now called **before** detaching the sandbox handle/route (while it does not mutate the backend). On allocation failure, image refs for `RuntimeImageOwner::PausedSandbox` are released and the state machine is rolled back from `Pausing` to `Running`.
- `run_lifecycle_operation` renamed to `run_cancellation_safe`; `replace_sandbox_network_policy` and `patch_sandbox_custom_extension_params` are now split into public wrapper + `_inner` implementation and routed through `run_cancellation_safe`, so a caller cancelling the request no longer aborts the update mid-flight — the background task still runs to completion.
- `build_user_egress_commands` now takes a `replace: bool` flag; when replacing, a `FlushChain` command is prepended to the same command batch sent to `apply_iptables_commands`, so flush + rule installation happen as a single atomic `iptables-restore` transaction. The standalone `flush_table_chain` helper was removed as it's no longer needed.

- Public API or generated protocol: N/A — no API surface changes.
- Configuration or defaults: N/A — no config changes.
- Snapshot manifest, artifact layout, or storage format: N/A — no format changes.
- Upgrade and rollback: No special steps; purely internal lifecycle/network behavior fixes.
- Host requirements, permissions, ports, or dependencies: N/A — no new host requirements.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [ ] Documentation updated

Commands and results:

```shell
make test-agent # All passed
```

Skipped checks and reasons:
- `services/` unchanged, so `make -C services test` skipped.

## Risks and reviewer notes

- Concurrent network policy/custom extension param updates are deliberately NOT serialized so that the orchestrator is kept simple.
- In-memory metadata update failures are deliberately IGNORED due to its rareness and the difficulty to implement a fully transactional architecture.

## Checklist

### DIFF
--- a/src/image/cache/mock.rs
+++ b/src/image/cache/mock.rs
@@ -1,0 +1,54 @@
+use anyhow::Result;
+use async_trait::async_trait;
+
+use super::{RuntimeImageOwner, RuntimeImageRefs};
+use crate::sandbox::RuntimeArtifactSet;
+use crate::types::SandboxId;
+
+#[derive(Debug, Default)]
+pub(crate) struct RecordingRuntimeImageRefs {
+    pinned: std::sync::Mutex<Vec<(RuntimeImageOwner, RuntimeArtifactSet)>>,
+    unpinned: std::sync::Mutex<Vec<RuntimeImageOwner>>,
+}
+
+impl RecordingRuntimeImageRefs {
+    pub(crate) fn pinned(&self) -> Vec<(RuntimeImageOwner, RuntimeArtifactSet)> {
+        self.pinned
+            .lock()
+            .expect("pinned refs mutex poisoned")
+            .clone()
+    }
+
+    pub(crate) fn unpinned(&self) -> Vec<RuntimeImageOwner> {
+        self.unpinned
+            .lock()
+            .expect("unpinned refs mutex poisoned")
+            .clone()
+    }
+}
+
+#[async_trait]
+impl RuntimeImageRefs for RecordingRuntimeImageRefs {
+    async fn pin(&self, owner: RuntimeImageOwner, artifacts: RuntimeArtifactSet) -> Result<()> {
+        self.pinned
+            .lock()
+            .expect("pinned refs mutex poisoned")
+            .push((owner, artifacts));
+        Ok(())
+    }
+
+    async fn unpin_best_effort(&self, owner: RuntimeImageOwner) {
+        self.unpinned
+            .lock()
+            .expect("unpinned refs mutex poisoned")
+            .push(owner);
+    }
+
+    async fn reconcile_paused(&self, _live_paused: &[SandboxId]) -> Result<()> {
+        Ok(())
+    }
+
+    async fn maintain_running(&self, _running: Vec<(SandboxId, RuntimeArtifactSet)>) -> Result<()> {
+        Ok(())
+    }
+}
--- a/src/image/cache/mod.rs
+++ b/src/image/cache/mod.rs
@@ -4,6 +4,9 @@ mod service;
 mod source_config;
 mod store;
 
+#[cfg(test)]
+mod mock;
+
 pub(crate) use store::{
     local_image_services_from_app_config, local_image_services_from_global_config,
     CachedImageConfig, OverlaybdLayerLocation, OverlaybdLayerStore, RuntimeImageOwner,
@@ -12,6 +15,7 @@ pub(crate) use store::{
 
 #[cfg(test)]
 pub(crate) mod test_support {
+    pub(crate) use super::mock::RecordingRuntimeImageRefs;
     pub(crate) use super::service::ImageCacheService;
     pub(crate) use super::store::test_local_image_services_from_service;
 }
--- a/src/orchestrator/service.rs
+++ b/src/orchestrator/service.rs
@@ -1045,6 +1045,27 @@ where
             return Err(error);
         }
 
+        // Allocate persistence space while the running handle and route are
+        // still attached. Allocation does not mutate the backend, so failure
+        // only needs to restore metadata and release the temporary image refs.
+        let artifact_root = match self.persister.allocate_artifact_root(&sandbox_id).await {
+            Ok(artifact_root) => artifact_root,
+            Err(err) => {
+                warn!(error = ?err, "failed to allocate paused sandbox artifact root");
+                self.release_image_refs(RuntimeImageOwner::PausedSandbox(sandbox_id))
+                    .await;
+                let _ = self
+                    .store
+                    .update_state_if_state(
+                        &sandbox_id,
+                        SandboxState::Running,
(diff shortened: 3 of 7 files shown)


guozy18 (approved):
The Image Cache handling looks good to me. If the remaining AI review comments have been addressed or confirmed as non-issues, this PR should be ready to merge.

PR #115, opened 2026-08-03
feat(cli): install shell completion loaders

rbalachandar:
Follow-up to #89, which adds the `aenv completion` command and visible aliases. This PR is intentionally scoped to automatic installation and removal.

## What it does

- Installs lightweight Bash, Zsh, and Fish completion loaders into standard completion directories.
- Uses the installed `aenv` binary path so loaders do not depend on an unrelated PATH binary.
- Canonicalizes prefix and HOME paths before selecting user versus system mode.
- Never modifies `.bashrc`, `.zshrc`, or Fish configuration files.
- Leaves unmanaged completion files and symlink destinations untouched.
- Atomically stages managed completion files in their destination directory and supports managed upgrades.
- `make uninstall-aenv` removes only files owned by the completion installer.
- Prints actionable Zsh `fpath` and `compinit` guidance for user-local installs.
- Treats staging and filesystem failures as non-fatal warnings so a CLI install is not aborted by completion setup.

The standalone installers remain self-contained because they are distributed as `curl | bash` entrypoints; their small loader-writing snippets mirror the shared Make/test helper without the previous shell configuration manager or synchronization machinery.

## Validation

- `bash -n` for all installer and test scripts
- ShellCheck for all installer and test scripts
- `bash scripts/tests/verify-completion-install.sh`
- `make check-completion-install`
- Functional coverage includes loader syntax, 0644 modes, install, managed upgrade, unmanaged Bash/Zsh/Fish preservation, symlink preservation, path traversal classification, staging-failure cleanup, standard system paths, and uninstall.

This draft should be reviewed after #89 is merged, since the loaders invoke the completion command introduced there.


LSX-s-Software:
Thanks for putting this together and for carefully covering installation, upgrades, removal, and existing-file handling.

After reconsidering the tradeoff, we don't plan to support automatic shell completion installation at this time. Correctly handling shell-specific paths, user versus system installations, upgrades, removal, and ownership of existing files introduces more installer complexity and maintenance burden than we'd like for this feature.

The existing `aenv completion <shell>` command and manual setup documentation should be sufficient for now. This also follows a common CLI pattern: tools usually generate completion scripts and leave persistent installation to users or package managers.

We'll close this PR as not planned for now. Thanks again for all the work you put into it.

PR #123, opened 2026-08-04
feat: add single-tenant API key authentication

yingdi-shan:
## Summary

- Require one shared single-tenant API key on runtime nodes and the gateway.
- Resolve the runtime API key during normal Rust server startup: `AENV_API_KEY`, then `/run/secrets/api-key`, then an atomically generated `$AENV_HOME/secrets/api-key`.
- Derive E2B-compatible per-sandbox credentials from the existing runtime seed:
  - `envdAccessToken = hex(HMAC-SHA256(seed, sandboxID))`
  - `trafficAccessToken = hex(HMAC-SHA256(seed, "sandbox-traffic-" + sandboxID))`
- Keep sandbox-token derivation and validation on runtime nodes. The gateway only checks that one sandbox credential is present before routing; the owning runtime performs the authoritative scoped validation.

## Credential boundaries

- `X-API-Key`: deployment control-plane credential shared by the gateway and runtime nodes.
- `e2b-traffic-access-token`: sandbox-scoped application proxy credential.
- `X-Access-Token`: secure envd credential, accepted only for the matching sandbox envd port.

The API key is independent from both sandbox credentials. Rotating an API key does not rotate sandbox tokens; rotating the runtime seed rotates both per-sandbox credentials.

Proxy routing strips AgentENV credentials before forwarding application requests and preserves application `Authorization`.

## Deployment behavior

- Native and standalone Docker starts generate and reuse managed API-key and sandbox-seed files during normal server startup.
- Docker Compose shares the runtime secrets directory between nodes and mounts it read-only on the gateway, which reads only the API key.
- Kubernetes stores the API key in `Secret/agentenv-auth`. The existing optional `agentenv-runtime-secrets/sandbox-access-token-hash-seed` contract is unchanged, avoiding seed migration during upgrades.
- Static multi-node deployments provide the API key to the gateway and all runtime nodes, and provide the sandbox seed only to runtime nodes.

## Out of scope

- External TLS termination is intentionally excluded and will be implemented separately.
- /metrics authentication is not supported
- https is not supported


sunkencity999:
Adding a production deployment data point, since this is a design-review vehicle and it may be useful to know how the current "no auth" state actually gets handled in the field.

We run AgentENV as the isolation layer under a multi-user internal agent platform — each end user gets one persistent sandbox, and an application-side shim maps users to sandbox IDs. Because there is no authentication today, we follow the README's guidance literally: the API is bound to `127.0.0.1:8000`, never exposed, and the only thing that talks to it is a local process. Everything reachable by a human sits behind a reverse proxy with its own authentication in front of the app, not in front of AgentENV.

Three observations that bear on the design:

**1. Loopback-only is a real deployment posture, not just a stopgap — please keep it viable.** For us the shared key would be defence in depth behind the loopback bind, not the primary boundary. If the implementation ends up *requiring* a key even for a loopback-only single-node install, that is a small friction on a config that is already safe; an "auth required unless bound to loopback" default, or simply generating the key automatically as you describe, avoids making people opt out of security to keep a working setup. The automatic generation in your Scope section reads like it already handles this well.

**2. The derived sandbox-scoped token is the part we would use most.** Our shim already knows which user owns which sandbox; AgentENV does not, and deliberately so. A token scoped to a sandbox ID means a compromised component can only reach the sandbox it was issued for, rather than the whole node. That is a meaningful reduction in blast radius for anyone running this multi-tenant on top, and it is the piece a shared key alone does not give you.

**3. Key rotation is the operational question we would ask first.** With a single shared key persisted at setup time, what does rotation look like for a running node with live sandboxes — is the expectation a restart, or can the key be re-read? If derived sandbox tokens are a function of the shared key, does rotating it invalidate every in-flight sandbox token? For long-lived sandboxes (ours persist across a working session and pause/resume rather than being recreated per command) that distinction matters quite a bit. Worth documenting whichever way it lands.

One smaller note: if the key ends up in a config file written by the setup paths, it would be helpful for the docs to state the expected file mode and owner explicitly. It is the kind of thing that is obvious to whoever writes it and non-obvious to whoever inherits the box.

Happy to test a branch against a real multi-user deployment if that is useful — we have a node running this pattern daily and can report back on anything that breaks under pause/resume or long-lived sandboxes.

src/api_key.rs
```diff
@@ -0,0 +1,195 @@
...
+
+    match temporary.persist_noclobber(path) {
+        Ok(_) => {
+            File::open(parent)?.sync_all()?;
+            info!(path = %path.display(), "generated managed API key");
+            Ok(key)
+        }
+        Err(error) if error.error.kind() == io::ErrorKind::AlreadyExists => {
+            read(path).context("load concurrently generated API key")
+        }
```
yingdi-shan: Fixed in d7d47f9. Managed API keys now use the shared managed-secret loader: it validates that the secrets directory is a non-symlink directory owned by the effective uid with mode 0700, opens the key with O_NOFOLLOW | O_NONBLOCK, and validates the opened object is a regular file owned by the effective uid with mode 0600 before reading it. The AlreadyExists fallback revalidates the directory and goes through the same no-follow open and file validation. External orchestrator-provided secrets remain unchanged.

src/sandbox/access.rs
```diff
@@ -562,16 +412,22 @@ mod tests {
...
         Ok(())
     }
+
+    #[cfg(unix)]
+    fn set_test_permissions(path: &Path, mode: u32) -> Result<()> {
+        use std::os::unix::fs::PermissionsExt;
+
+        fs::set_permissions(path, fs::Permissions::from_mode(mode))?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. Added a cfg(not(unix)) no-op test helper so the access tests compile on non-Unix targets.

deploy/k8s/run.sh
```diff
@@ -10,6 +10,7 @@ MODE="$1"
 shift
 KUBECTL_BIN="${KUBECTL:-kubectl}"
 OVERLAY_NAME="${K8S_OVERLAY:-default}"
+NAMESPACE="${K8S_NAMESPACE:-agentenv-system}"
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. The temporary Kustomization, Namespace resource, and scheduler discovery config now use the validated K8S_NAMESPACE value, so preflight, generated resources, and scheduler registration agree.

src/api/proxy.rs
```diff
@@ -142,6 +148,65 @@ where
...
+    match parse_host_proxy_route(request_host(request), domains) {
+        Ok(Some(route)) => return Some((route.sandbox_id, route.target_port)),
+        Err(_) => return None,
+        Ok(None) => {}
+    }
+
+    Some((
+        parse_sandbox_id_header(request.headers()).ok()?,
+        parse_target_port_header(request.headers()).ok()?,
+    ))
```
yingdi-shan: Fixed in amended commit b6bedef. Host-route parse failures and absent host routes now fall back to explicit routing headers, while valid host routes retain precedence for host-only requests.

src/api/impls/auth.rs
```diff
@@ -1,41 +1,100 @@
...
+    if request.uri().path() == "/health" && !proxy_request {
+        return next.run(request).await;
+    }
+
+    let api_impl = api_impl.as_ref();
+    let mut authorized = api_impl.has_valid_api_key(request.headers());
+    if !authorized && proxy_request {
+        if let Some((sandbox_id, target_port)) =
+            proxy::route_for_auth(&request, api_impl.sandbox_proxy_domains())
+        {
```
yingdi-shan: Fixed in amended commit b6bedef. Explicit /proxy routes now derive the auth identity from routing headers, matching proxy dispatch; host routing is used only for non-explicit routes.

src/api/impls/auth.rs
```diff
@@ -1,41 +1,100 @@
...
+fn single_header<'a>(headers: &'a HeaderMap, name: &str) -> Option<&'a HeaderValue> {
+    let mut values = headers.get_all(name).iter();
+    let value = values.next()?;
+    values.next().is_none().then_some(value)
+}
+
+impl ApiImpl {
+    pub(crate) fn has_valid_api_key(&self, headers: &HeaderMap) -> bool {
+        single_header(headers, API_KEY_HEADER)
+            .is_some_and(|value| value.as_bytes() == self.api_key.as_bytes())
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. API-key comparison now enforces equal length and uses subtle constant-time equality.

src/api/proxy.rs
```diff
@@ -963,19 +1045,24 @@ fn build_upstream_uri_with_scheme(
...
     headers.remove(SANDBOX_ID_HEADER);
     headers.remove(E2B_SANDBOX_ID_HEADER);
     headers.remove(TARGET_PORT_HEADER);
     headers.remove(E2B_TARGET_PORT_HEADER);
+    headers.remove(API_KEY_HEADER);
+    headers.remove(TRAFFIC_ACCESS_TOKEN_HEADER);
     headers.remove(header::HOST);
+    if target_port != envd_port {
+        headers.remove(ENVD_ACCESS_TOKEN_HEADER);
+    }
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. x-access-token is independently validated for the resolved sandbox and port, and is stripped before forwarding unless that validation succeeds, including API-key-authenticated requests.

src/api/proxy.rs
```diff
@@ -735,6 +807,14 @@ async fn resolve_proxy_request(
         }
     };
 
+    let metadata = api_impl
+        .orchestrator()
+        .get_sandbox(&sandbox_id)
+        .await
+        .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR.into_response())?
+        .ok_or_else(|| proxy_error_response(&ProxyRequestError::SandboxNotFound(sandbox_id)))?;
+    let envd_port = effective_envd_port(&metadata);
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. Removed the redundant metadata-store read from proxy resolution; metadata is read only when validating an x-access-token or when auto-resume needs it.

src/managed_secret.rs
```diff
@@ -0,0 +1,214 @@
...
+    create_directory(parent)?;
+    validate_directory_identity(parent).with_context(|| {
+        format!(
+            "validate managed secret directory ownership {}",
+            parent.display()
+        )
+    })?;
+    set_permissions(parent, 0o700)?;
+    validate_directory(parent)
+        .with_context(|| format!("validate managed secret directory {}", parent.display()))?;
```
yingdi-shan, line 93: Reviewed for amended commit b6bedef. I left the pathname TOCTOU suggestion unchanged: the managed directory must be a non-symlink directory owned by the effective UID with mode 0700, so replacing it requires the same UID or privileged mount capability and already provides equivalent secret access. An openat-based rewrite would add substantial platform-specific complexity without changing that threat model.

src/managed_secret.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::OpenOptionsExt;
+
+        options.custom_flags(libc::O_NOFOLLOW | libc::O_NONBLOCK);
+    }
+
+    options.open(path)
+}
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. Managed-secret creation and opening now fail closed on non-Unix platforms because the required no-follow file semantics are unavailable.

src/api_key.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+fn read(path: &Path) -> Result<String, io::Error> {
+    let value = fs::read_to_string(path)?;
+    validate(&value).map_err(io::Error::other)
+}
+
+fn validate(value: &str) -> Result<String> {
+    let value = value.trim();
+    if value.len() < 32
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. API keys from environment, external secrets, and managed storage now share the 32 to 4096 byte bound before acceptance.

scripts/tests/e2e/suites/09_e2b_compat.sh
```diff
@@ -10,8 +10,7 @@ log "Suite: E2B Compatibility"
 
 export E2B_API_URL="${AENV_URL}"
 export E2B_SANDBOX_URL="${AENV_PROXY_URL}"
-export E2B_API_KEY="e2b_000000"
-export E2B_ACCESS_TOKEN="${AENV_API_KEY}"
+export E2B_API_KEY="${AENV_API_KEY}"
```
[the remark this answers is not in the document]
yingdi-shan: Fixed in amended commit b6bedef. The E2B compatibility suite now explicitly unsets E2B_ACCESS_TOKEN after setting E2B_API_KEY.

yingdi-shan:
Review follow-up for amended commit b6bedef: external /run/secrets/api-key loading now uses the same bounded 4096-byte limit as managed keys, reading at most one byte beyond the limit before rejecting oversized input. The amended commit also fixes the Docker Compose named-volume initialization, auth/proxy route consistency, token forwarding, constant-time API-key comparison, namespace rendering, and E2B token isolation. Local fmt, clippy, and unit tests pass; the prior integration failure was only the CI runner timing out on www.baidu.com:443.

docs/src/security/authentication.md
LSX-s-Software, line 1: I think we should move this file to `docs/src/security/`.

docs/src/integration/e2b.md
```diff
@@ -12,10 +12,17 @@ Set environment variables to point at your AgentENV server. See [Environment Var
 # Single-node example
 export E2B_API_URL=http://127.0.0.1:8000
 export E2B_SANDBOX_URL=${E2B_API_URL}
-export E2B_API_KEY=e2b_000000
-export E2B_ACCESS_TOKEN=dummy
+export E2B_API_KEY=${AENV_API_KEY}
 ```
 
+No `E2B_ACCESS_TOKEN` is needed. AgentENV returns `trafficAccessToken` when
```
LSX-s-Software: We should remove the "No `E2B_ACCESS_TOKEN` is needed." and "This is transport data, not the deprecated user-supplied `E2B_ACCESS_TOKEN`."

services/gateway/cmd/main.go
```diff
@@ -22,13 +25,85 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	apiKeyEnv         = "AENV_API_KEY"
+	defaultAPIKeyPath = "/run/secrets/api-key"
+	maxAPIKeyLen      = 4096
```
LSX-s-Software: I think this is too long for a reasonable API key.

scripts/tests/e2e/suites/09_e2b_compat.sh
```diff
@@ -10,8 +10,8 @@ log "Suite: E2B Compatibility"
...
+unset E2B_ACCESS_TOKEN
```
LSX-s-Software: No one has set this variable, so we don't need to unset it here.

scripts/tests/e2e/lib/helpers.sh
```diff
@@ -121,44 +121,6 @@ if [[ -z "${E2E_HELPERS_SH_LOADED:-}" ]]; then
...
-    [[ -z "$_E2E_HEADERS" ]] && _E2E_HEADERS=$(mktemp)
-    curl -s -o "$_E2E_BODY" -D "$_E2E_HEADERS" -w '%{http_code}' \
-      -H "X-Admin-Token: ${AENV_ADMIN_TOKEN}" \
-      "${AENV_URL}${path}" > "$_E2E_STATUS" 2>/dev/null || true
-    HTTP_STATUS=$(<"$_E2E_STATUS")
-    HTTP_BODY=$(<"$_E2E_BODY")
-    HTTP_HEADERS=$(<"$_E2E_HEADERS")
-  }
-
-  api_admin_get_with_headers_at() {
```
LSX-s-Software: We'd better keep these helpers, because we'll definitely need them later.

src/api/openapi.yml
```diff
@@ -302,7 +302,7 @@ components:
         allowPublicTraffic:
           type: boolean
           default: true
-          description: Specify if the sandbox URLs should be accessible only with authentication.
+          description: Specify if the sandbox URLs should be accessible without a traffic access token.
```
LSX-s-Software: The original description was directly copied from E2B's openapi spec, so I think we don't need to modify it.

src/api_key.rs
```diff
@@ -0,0 +1,209 @@
...
+use rand::{rngs::SysRng, TryRng};
+use tracing::info;
+
+use crate::cfg::AppConfig;
+use crate::managed_secret::{self, CreateOutcome};
+
+const API_KEY_ENV: &str = "AENV_API_KEY";
+const EXTERNAL_API_KEY_PATH: &str = "/run/secrets/api-key";
+const MANAGED_API_KEY_RELATIVE_PATH: &str = "secrets/api-key";
+const API_KEY_MAX_LEN: usize = 4096;
```
LSX-s-Software: I think this is too long.

src/api/proxy.rs
```diff
@@ -358,6 +404,14 @@ fn has_routing_header(headers: &HeaderMap) -> bool {
     headers.get(SANDBOX_ID_HEADER).is_some() || headers.get(E2B_SANDBOX_ID_HEADER).is_some()
 }
 
+pub(crate) fn effective_envd_port(metadata: &SandboxMetadata) -> u16 {
+    metadata
+        .paused_state
+        .as_ref()
+        .and_then(|state| state.control_plane_port())
+        .unwrap_or_else(|| ConfigManager::global_config().tools.control_plane_port)
```
LSX-s-Software, line 412: This is a temporary workaround to get the actual control plane port. We should record it in `SandboxMetadata` in the future.

src/sandbox/network/policy.rs
```diff
@@ -73,20 +73,42 @@ impl SandboxNetworkEgressPolicy {
...
+            allow_public_traffic: true,
             base_policy,
             egress,
         }
     }
 
+    pub fn with_allow_public_traffic(mut self, allow_public_traffic: bool) -> Self {
+        self.allow_public_traffic = allow_public_traffic;
+        self
+    }
```
LSX-s-Software: We should not use an implicit `allow_public_traffic: true` in the `new()` constructor and then use `with_allow_public_traffic` to override it. The correct way is to add `allow_public_traffic` to the param list of `new()` and remove the `with_allow_public_traffic()`.

src/api_key.rs
LSX-s-Software, line 1: I think we can encapsulate the API key logic in a wrapper type like `struct ApiKey(String)`, so that callers can use OOP-style methods like `ApiKey::resolve()` instead of calling free functions.

LSX-s-Software (changes requested):

LSX-s-Software (approved):

PR #127, opened 2026-08-05
feat(sandbox): add E2B-compatible secure sandbox support

LSX-s-Software:
## What

Introduce end-to-end support for E2B-compatible secure sandboxes. When a sandbox is created with `secure: true`, AgentENV generates a sandbox-specific envd access token and enforces it across the complete sandbox lifecycle.

- Add the optional `secure` sandbox create field and return `envdAccessToken` from create, get, connect, and fork responses.
- Pass secure tokens through Firecracker MMDS and envd initialization, and attach them to internal HTTP, Connect, aenv CLI, and E2B SDK requests.
- Give secure fork children independent tokens and preserve token identity across pause, server restart, and resume.
- Persist the sandbox's secure mode while keeping legacy records non-secure by default.
- Manage a persistent node-local token seed automatically for single-node installations, while allowing an explicit shared seed for multi-node deployments.
- Document deployment secret requirements and exercise secure mode explicitly in the E2B Python, TypeScript, and code-interpreter compatibility suites.

## Why

AgentENV did not previously implement the E2B `secure` sandbox contract. Current E2B SDKs create secure sandboxes by default, so compatibility requires AgentENV to return an envd access token and enforce it transparently for SDK and CLI operations.

Introducing secure support should not force existing or single-node users to add configuration before upgrading. AgentENV therefore creates and reuses a managed local seed when no explicit seed is supplied, while multi-node operators can configure one shared seed through their deployment secret mechanism.

## Related issue

N/A - no linked issue.

## Scope and non-goals

In E2B semantics, `secure` authenticates envd control communication. This PR does not add authentication to arbitrary application ports or implement cross-node sandbox migration. The install script remains unchanged.

- Tokens are HMAC-SHA256 values derived from the effective seed and sandbox ID. The token itself is not persisted in sandbox metadata.
- Firecracker MMDS stores only the token hash. `/init` receives the token in both `X-Access-Token` and the request body, matching envd's contract.
- Internal envd clients and proxy routes supply or validate the token as appropriate.
- Fork children derive tokens from their own sandbox IDs and cannot authenticate with the source sandbox's token.
- Secure state is persisted with a serde default of `false`, so legacy sandbox records continue to restore as non-secure.
- Explicit environment or TOML configuration takes precedence over managed state. Changing the effective seed rotates existing secure sandbox tokens.
- Without an explicit seed, normal startup reads or atomically creates `$AENV_HOME/secrets/sandbox-access-token-hash-seed`. The directory is `0700`, the file is `0600`, and malformed, unreadable, symlinked, or unexpectedly missing state fails instead of rotating silently.
- Concurrent startup uses no-clobber persistence and all contenders reuse the winning seed.

- Public API or generated protocol: Introduces optional `secure` request fields and optional `envdAccessToken` response fields. Omitting `secure` preserves the previous non-secure behavior.
- Configuration or defaults: Adds optional `[sandbox].access_token_hash_seed` / `AENV_SANDBOX_ACCESS_TOKEN_HASH_SEED`; single-node startup auto-generates a persistent seed when unset.
- Snapshot manifest, artifact layout, or storage format: No snapshot format change. Persisted sandbox metadata adds `secure` with a legacy default of `false`.
- Upgrade and rollback: Existing users can upgrade without changing configuration. Preserve `$AENV_HOME/secrets/sandbox-access-token-hash-seed` once secure sandboxes exist; deleting or changing it invalidates their tokens.
- Host requirements, permissions, ports, or dependencies: No new ports or host packages. Kubernetes runtime nodes require the documented `agentenv-runtime-secrets` Secret; the local-dev overlay provides a test-only value.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] Documentation updated

Skipped checks and reasons:

- `make -C services test` was not run because `services/` is unchanged.
- Benchmarks were not run because this change affects startup secret handling and request authentication rather than snapshot or steady-state data paths.

## Risks and reviewer notes

- Review the API-to-envd token flow as one new end-to-end secure sandbox contract rather than as an isolated seed-management change.
- Review `src/sandbox/access.rs` for managed-secret durability, permissions, and concurrent startup behavior.
- Review fork and resume identity replacement to ensure a child or resumed runtime cannot retain stale token metadata.
- Multi-node deployments must use the same explicit seed on every runtime node before cross-node recovery of a sandbox ID is introduced.

## Checklist

### DIFF
--- a/docs/src/deployment/docker-compose.md
+++ b/docs/src/deployment/docker-compose.md
@@ -32,11 +32,38 @@ git clone https://github.com/kvcache-ai/AgentENV.git
 cd AgentENV
 ```
 
+## Configure the Shared Access-Token Seed
+
+Multi-node deployments should explicitly configure one envd access-token seed
+and use the same value on every **runtime node**. Do not rely on the node-local
+auto-generated seed in a scheduler-managed deployment.
+
+Generate the value once and store it in your secret manager:
+
+```bash
+openssl rand -hex 32
+```
+
+For the checked-in Compose stack, create a private configuration copy outside
+the repository and set the generated value in it:
+
+```bash
+install -d -m 0700 "$HOME/.config/agentenv"
+install -m 0600 config/default.toml "$HOME/.config/agentenv/cluster.toml"
+```
+
+```toml
+[sandbox]
+access_token_hash_seed = "<shared-secret>"
+```
+
+Changing this value rotates the access tokens for existing secure sandboxes.
+
 ## Start the Cluster
 
 ```bash
 sudo bash scripts/docker-setup.sh
-make deploy-up
+CONFIG_PATH="$HOME/.config/agentenv/cluster.toml" make deploy-up
 ```
 
 To enable host-based sandbox data-plane URLs, set the shared sandbox proxy
@@ -44,7 +71,7 @@ domain variable when starting the stack:
 
 ```bash
 SANDBOX_PROXY_DOMAINS=sandbox.example.com \
-make deploy-up
+CONFIG_PATH="$HOME/.config/agentenv/cluster.toml" make deploy-up
 ```
 
 Compose passes this value to both the gateway routing allowlist and runtime
--- a/docs/src/deployment/docker-compose.md
+++ b/docs/src/deployment/docker-compose.md
@@ -32,38 +32,16 @@ git clone https://github.com/kvcache-ai/AgentENV.git
 cd AgentENV
 ```
 
-## Configure the Shared Access-Token Seed
+## Configure the Access-Token Seed (Optional)
 
-Multi-node deployments should explicitly configure one envd access-token seed
-and use the same value on every **runtime node**. Do not rely on the node-local
-auto-generated seed in a scheduler-managed deployment.
-
-Generate the value once and store it in your secret manager:
-
-```bash
-openssl rand -hex 32
-```
-
-For the checked-in Compose stack, create a private configuration copy outside
-the repository and set the generated value in it:
-
-```bash
-install -d -m 0700 "$HOME/.config/agentenv"
-install -m 0600 config/default.toml "$HOME/.config/agentenv/cluster.toml"
-```
-
-```toml
-[sandbox]
-access_token_hash_seed = "<shared-secret>"
-```
-
-Changing this value rotates the access tokens for existing secure sandboxes.
+See [Secure Sandboxes](../security/secure-sandboxes.md)
+if the deployment needs future cross-node sandbox recovery.
 
 ## Start the Cluster
 
 ```bash
 sudo bash scripts/docker-setup.sh
-CONFIG_PATH="$HOME/.config/agentenv/cluster.toml" make deploy-up
+make deploy-up
 ```
 
 To enable host-based sandbox data-plane URLs, set the shared sandbox proxy
@@ -71,7 +49,7 @@ domain variable when starting the stack:
 
 ```bash
 SANDBOX_PROXY_DOMAINS=sandbox.example.com \
-CONFIG_PATH="$HOME/.config/agentenv/cluster.toml" make deploy-up
+make deploy-up
 ```
 
 Compose passes this value to both the gateway routing allowlist and runtime
--- a/docs/src/deployment/static-multi-node.md
+++ b/docs/src/deployment/static-multi-node.md
@@ -62,11 +62,13 @@ curl -fsSL https://raw.githubusercontent.com/kvcache-ai/AgentENV/main/scripts/in
 ```
 
 Edit `/etc/default/aenv` on each machine without removing the paths written by
-the installer. Node A uses:
+the installer. Generate one access-token seed with `openssl rand -hex 32`, store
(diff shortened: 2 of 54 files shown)


yingdi-shan:
We should add HTTPS and API-KEY authentication before merging this PR — without them, an access token alone doesn't provide sufficient security. PR #123 provides a basic draft.

LSX-s-Software:
> We should add HTTPS and API-KEY authentication before merging this PR — without them, an access token alone doesn't provide sufficient security. PR #123 provides a basic draft.

I agree that TLS is required when AgentENV is exposed over an untrusted network, but I don’t think #123 should block this PR.

#123 is still a draft, and **its token model currently has several unresolved issues**:

- It derives one token and returns it as both `envdAccessToken` and `trafficAccessToken`, but E2B defines them as separate credentials for different trust boundaries, which is more reasonable.
- It populates `envdAccessToken` in the API response without installing that token into envd through MMDS/init. As a result, the SDK receives a token, **but envd is not actually secured by it**.
- It accepts `X-Access-Token` for arbitrary sandbox proxy traffic and forwards that credential to the upstream sandbox application. In E2B, `X-Access-Token` is for envd communication, while proxy traffic uses the separate `e2b-traffic-access-token` header.
- It also explicitly excludes HTTPS/TLS from its scope, so **merging #123 would not satisfy the proposed HTTPS prerequisite anyway**.

These are separate authentication layers and should remain separate:

1. This PR, which implements E2B-compatible `secure` semantics and per-sandbox envd authentication.
2. API-key authentication protects the AgentENV control-plane API.
3. Traffic-token authentication protects proxied sandbox ports.
4. TLS protects all of these credentials in transit.

The safer ordering is to merge the correct envd authentication contract in this PR first, then rebase #123 and implement API-key and traffic authentication on top without conflating the credentials. TLS can be handled independently at the deployment or ingress layer.

yingdi-shan (changes requested):
As we discussed before, I think we should avoid adding a shared seed in this PR. It only serves as a workaround for VM migration, which isn't on the near-term roadmap. Removing it here would simplify deployment.

LSX-s-Software:
The seed is already **OPTIONAL**. We are not requiring users to configure it or enabling cross-node migration in this PR.

This only documents an optional way for operators to prepare for future multi-node sandbox recovery, so they won’t need to reconfigure the entire cluster later. In practice, configuring the seed takes only a few commands, which is negligible compared with the setup already required for a multi-node deployment.

Users who do not need this can simply leave the seed unset and use the automatically managed node-local seed.

yingdi-shan:
> The seed is already **OPTIONAL**. We are not requiring users to configure it or enabling cross-node migration in this PR.
> 
> This only documents an optional way for operators to prepare for future multi-node sandbox recovery, so they won’t need to reconfigure the entire cluster later. In practice, configuring the seed takes only a few commands, which is negligible compared with the setup already required for a multi-node deployment.
> 
> Users who do not need this can simply leave the seed unset and use the automatically managed node-local seed.

The docs does not say that. Maybe you should update the docs.

docs/src/deployment/docker-compose.md
```diff
@@ -32,19 +32,46 @@ git clone https://github.com/kvcache-ai/AgentENV.git
 cd AgentENV
 ```
 
+## Configure the Shared Access-Token Seed (Recommended but Optional)
```
yingdi-shan: Optional is enough, using a shared seed is not recommended because it impacts security.

docs/src/deployment/docker-compose.md
```diff
@@ -32,19 +32,46 @@ git clone https://github.com/kvcache-ai/AgentENV.git
...
+```
+
+Changing this value rotates the access tokens for existing secure sandboxes.
+
 ## Start the Cluster
 
 ```bash
 sudo bash scripts/docker-setup.sh
-make deploy-up
+CONFIG_PATH="$HOME/.config/agentenv/cluster.toml" make deploy-up
```
yingdi-shan: We should remove the CONFIG_PATH here, since the previos step is optional.

docs/src/deployment/static-multi-node.md
```diff
@@ -62,11 +62,16 @@ curl -fsSL https://raw.githubusercontent.com/kvcache-ai/AgentENV/main/scripts/in
 ```
 
 Edit `/etc/default/aenv` on each machine without removing the paths written by
-the installer. Node A uses:
+the installer.
+
+Configuring one shared access-token seed with `openssl rand -hex 32` is recommended, but optional. It prepares the cluster for future recovery of the same sandbox on another node without requiring a later cluster-wide configuration change. If you do not need that, omit the seed and AgentENV will create a persistent node-local seed automatically on each node. When using an explicit shared seed, store it securely and set the same value on every runtime node.
```
yingdi-shan: We are saying this over and over again. Maybe we should move all the sharing seed configuration into a separate markdown? The deployment path should remain as simple as possible.

yingdi-shan (changes requested):

yingdi-shan (approved):

PR #86, opened 2026-07-30
feat: add ARM64 support for AgentENV

zlzGithub-0801:
## What

Add ARM64/aarch64 support across AgentENV runtime dependency provisioning, release artifacts, and Docker images.

## Why

AgentENV installation and Server releases previously assumed x86_64. On ARM64 hosts, the installer rejected the architecture and dependency setup could not resolve an architecture-specific kernel artifact.

## Related issue

Closes #60, #63 

## Scope and non-goals

Included:

- Firecracker and Linux kernel download paths
- Release behavior
- Installation script

Not included:

- Other AgentENV runtime logic

## Design and behavior changes

- During setup, download the matching Firecracker binary and Linux kernel based on the configured version and host architecture.
- Add `aenv-server-linux-aarch64.tar.gz` to the release artifacts and publish the Docker image as a multi-architecture image.
- Update the installation script to select the appropriate release artifact based on the host architecture.

## Compatibility and operations

- Public API or generated protocol: N/A — no API schema or generated code changes.
- Configuration or defaults: The bundled Firecracker and kernel URLs now point to the `kvcache-ai/firecracker` GitHub Release and include architecture-specific asset names.
- Snapshot manifest, artifact layout, or storage format: N/A — no snapshot or persistent storage format changes.
- Upgrade and rollback: N/A — Existing x86_64 installations remain supported.
- Host requirements, permissions, ports, or dependencies: ARM64 execution requires an aarch64 Linux host with KVM access and the same existing AgentENV host requirements.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] Relevant Rust integration tests
- [ ] `make -C services test` (required when `services/` changes)
- [x] Generated clients/server regenerated with the documented `make` target
- [x] Documentation updated
- [ ] Benchmarks or performance comparison completed

Commands and results:

```text

```

Skipped checks and reasons:

This PR does not change `services/` and doesn't have performance-sensitive change either.

## Checklist

- [x] The PR contains one coherent change and no unrelated formatting or refactoring.
- [x] New behavior is covered by tests, or I explained why testing is impractical.
- [x] Logs and examples contain no credentials, tokens, or private registry information.
- [x] I did not manually edit generated code without updating its source and regenerating it.

PR #129, opened 2026-08-06
feat: add openEuler support to Linux distro detection

liunianxuxie:
## What

Add openEuler support to AgentENV's Linux distribution detection so --setup-only works on openEuler hosts.

## Why

openEuler is a glibc-based RPM-family Linux distro widely used on ARM (Kunpeng) servers in China. The upstream ARM64 support (#86) covers Ubuntu/Debian but fails on openEuler with two errors:

1. --setup-only fails with unsupported Linux distribution — distro_from_id in src/setup/packages.rs does not recognize openEuler, so the runtime package list cannot be resolved.
2. overlaybd provisioning fails with unsupported overlaybd release target — configured_overlaybd_release in src/setup/overlaybd.rs does not map openEuler to the RPM-family fallback.

## Related issue

No issue filed

## Scope and non-goals

Included:

- openEuler (ID=openEuler, lowercased to openeuler by the parser) recognized as RPM-family in src/setup/packages.rs
- openEuler mapped to the existing Ubuntu 22.04 overlaybd fallback in src/setup/overlaybd.rs
- Test coverage for both code paths

Excluded:

- No changes to dependency URLs, manifests, or tooling (upstream aenv-deps release already provides aarch64 binaries)
- No changes to tools-image/ build, network mirrors, or documentation
- No changes to non-RPM distributions

openEuler already matches the existing RPM-family pattern used by centos/rhel/tencentos:

- src/setup/packages.rs: distro_from_id maps openeuler/openEuler to Distro::Centos, reusing the existing dnf/yum package-manager detection and the libaio runtime package list.
- src/setup/overlaybd.rs: configured_overlaybd_release maps openEuler to ubuntu1.22.04.{arch} — the oldest published Ubuntu overlaybd asset, chosen because it bundles its own shared libraries and runs on any glibc-based distro (same reasoning as the existing tencentos/centos/rhel fallback).

- Public API or generated protocol: N/A
- Configuration or defaults: N/A
- Snapshot manifest, artifact layout, or storage format: N/A
- Upgrade and rollback: N/A — additive recognition of a previously-unsupported OS
- Host requirements, permissions, ports, or dependencies: openEuler hosts additionally need dpkg (to extract the overlaybd .deb)

## Validation

- [ ] `make fmt`
- [ ] `make clippy`
- [ ] `make test-unit`
- [ ] Documentation updated

Commands and results:

```text

```

Skipped checks and reasons:

## Checklist

- [ ] The PR contains one coherent change and no unrelated formatting or refactoring.
- [ ] New behavior is covered by tests, or I explained why testing is impractical.
- [ ] Logs and examples contain no credentials, tokens, or private registry information.
- [ ] I did not manually edit generated code without updating its source and regenerating it.

### DIFF
--- a/src/setup/overlaybd.rs
+++ b/src/setup/overlaybd.rs
@@ -136,8 +136,14 @@ fn configured_overlaybd_release(
         // so it runs fine on other glibc-based distros. Fall back to the
         // oldest published Ubuntu build (lowest glibc requirement) for known
         // RPM-family distros so the CLI tools install without a native asset.
-        "tencentos" | "centos" | "centos-stream" | "rhel" | "redhat"
-        | "redhatenterpriseserver" | "openeuler" | "openEuler" => {
+        "tencentos"
+        | "centos"
+        | "centos-stream"
+        | "rhel"
+        | "redhat"
+        | "redhatenterpriseserver"
+        | "openeuler"
+        | "openEuler" => {
             format!("ubuntu1.{}.{}", FALLBACK_UBUNTU_VERSION, target.arch)
         }
         other => bail!(
@@ -577,7 +583,14 @@ mod tests {
             ),
         };
 
-        for os_id in ["centos", "centos-stream", "tencentos", "rhel", "openeuler", "openEuler"] {
+        for os_id in [
+            "centos",
+            "centos-stream",
+            "tencentos",
+            "rhel",
+            "openeuler",
+            "openEuler",
+        ] {
             let release = configured_overlaybd_release(
                 &config,
                 &OverlaybdReleaseTarget {
--- a/src/setup/overlaybd.rs
+++ b/src/setup/overlaybd.rs
@@ -136,7 +136,8 @@ fn configured_overlaybd_release(
         // so it runs fine on other glibc-based distros. Fall back to the
         // oldest published Ubuntu build (lowest glibc requirement) for known
         // RPM-family distros so the CLI tools install without a native asset.
-        "tencentos" | "centos" | "centos-stream" | "rhel" | "redhat" | "redhatenterpriseserver" => {
+        "tencentos" | "centos" | "centos-stream" | "rhel" | "redhat"
+        | "redhatenterpriseserver" | "openeuler" | "openEuler" => {
             format!("ubuntu1.{}.{}", FALLBACK_UBUNTU_VERSION, target.arch)
         }
         other => bail!(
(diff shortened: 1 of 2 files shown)


liunianxuxie:
@yingdi-shan Could you please take a look and merge this PR to enable openEuler support?

yingdi-shan:
> @yingdi-shan Could you please take a look and merge this PR to enable openEuler support?

LGTM. Please fix the fmt error.

zdy782:
Could you share a bit more detail on the host setup you tested with? Specifically:

- Which openEuler version and kernel are you running on (e.g. `uname -r`)?
- README lists **Linux kernel 6.8+** as a prerequisite, mainly because the optional zero-copy path (`UBLK_F_AUTO_BUF_REG`) requires kernel 6.8+. If your openEuler host ships an older kernel (e.g. 6.6), how did you handle that — did you upgrade the kernel, or are you relying on the `ublk_drv` module being available on 6.6 with everything else working?

yingdi-shan (approved):

liunianxuxie:
> Could you share a bit more detail on the host setup you tested with? Specifically:
> 
> * Which openEuler version and kernel are you running on (e.g. `uname -r`)?
> * README lists **Linux kernel 6.8+** as a prerequisite, mainly because the optional zero-copy path (`UBLK_F_AUTO_BUF_REG`) requires kernel 6.8+. If your openEuler host ships an older kernel (e.g. 6.6), how did you handle that — did you upgrade the kernel, or are you relying on the `ublk_drv` module being available on 6.6 with everything else working?

I tested on openEuler 24.03 SP3 with the following kernel version:
6.6.0-132.0.0.111.oe2403sp3.aarch64

I noticed that openEuler does not enable ublk by default, so I rebuilt the kernel from source and enabled ublk support.

zdy782:
> > Could you share a bit more detail on the host setup you tested with? Specifically:
> > 
> > * Which openEuler version and kernel are you running on (e.g. `uname -r`)?
> > * README lists **Linux kernel 6.8+** as a prerequisite, mainly because the optional zero-copy path (`UBLK_F_AUTO_BUF_REG`) requires kernel 6.8+. If your openEuler host ships an older kernel (e.g. 6.6), how did you handle that — did you upgrade the kernel, or are you relying on the `ublk_drv` module being available on 6.6 with everything else working?
> 
> I tested on openEuler 24.03 SP3 with the following kernel version: 6.6.0-132.0.0.111.oe2403sp3.aarch64
> 
> I noticed that openEuler does not enable ublk by default, so I rebuilt the kernel from source and enabled ublk support.

Thanks for the clarification!

Just to confirm: since you built and tested this on Linux 6.6, does that mean the zero-copy path using UBLK_F_AUTO_BUF_REG was not enabled/used in your setup?

If so, would disabling this zero-copy path have any noticeable impact on performance, such as throughput, latency, or CPU overhead, compared with running on a 6.8+ kernel with UBLK_F_AUTO_BUF_REG enabled?

Thanks again for your help!

Issue #120, opened 2026-08-04
Advance memory snapshot epochs for linear incremental OverlayBD chains


xeonliu:
### Affected component

Snapshot / template builder

### Problem statement

AgentENV can capture a running sandbox without terminating it. The Firecracker
backend pauses the VM, creates snapshot artifacts, and then resumes the same
Firecracker process in place. However, the sandbox's launch configuration
continues to refer to the memory image from which that process originally
started.

This is compatible with the current `mincore` behavior because each capture
can include a broad resident-page set and remain a sibling over the original
launch parent. It is not sufficient for true per-checkpoint dirty-page
snapshots.

For example, suppose a VM starts from `S0`:

```text
S0
  -> write page set A
  -> capture S1
  -> resume the same process
  -> write page set B
  -> capture S2
```

If the dirty epoch is reset after `S1`, the newest layer for `S2` naturally
contains B, not A. If AgentENV still builds `S2` over `S0`, the result is
incomplete:

```text
incorrect: S2 = S0 + delta(B)
required:  S2 = S0 + delta(A) + delta(B)
```

Simply enabling dirty tracking and clearing its bitmap after every capture can
therefore create snapshots that are missing memory changes. AgentENV needs an
explicit live memory parent, a transactional dirty epoch, and artifact
lifetime rules before it can provide deltas relative to the latest successful
capture.

Companion issue: #119 

### Use case

Agent and RL workloads frequently capture or fork the same running sandbox at
multiple decision points. A long-running VM may modify only a small page set
between checkpoints, while each checkpoint becomes a base for later rollback,
fork, template publication, or migration.

These workloads need a linear memory checkpoint chain in which each new
OverlayBD layer is relative to the immediately preceding successful local
capture. Without that chain, dirty tracking must either retain an ever-growing
write set relative to the original launch parent or risk producing an
incomplete snapshot.

### Current behavior and workarounds

During capture, `snapshot_to_dir` chooses inherited memory layers from
`LaunchMode::Resume` and appends the new `overlaybd.commit`. A non-terminal
`capture_snapshot` then resumes the same process, but it does not advance the
launch mode or another explicit current-memory-parent field to the newly
captured `mem_image.json`.

Current options are insufficient:

- Keep dirty pages accumulated relative to the original launch parent. This is
  correct but later captures repeat every page written during the VM's
  lifetime.
- Use `mincore` so captures contain a broad resident set. This avoids an epoch
[...]

PR #134, opened 2026-08-06
feat: add opt-in dirty-page tracking for memory snapshots

### LINKED ISSUE #119: Add opt-in dirty-page tracking for memory snapshot page selection
### Affected component

Firecracker runtime

### Problem statement

AgentENV currently configures fresh VMs with `track_dirty_pages = false`, and file-backed snapshot restore does not set `SnapshotLoadParams.track_dirty_pages`. Memory snapshots are requested as Firecracker `Diff` snapshots, but without dirty tracking Firecracker uses `mincore(2)` to select pages.

That makes the result sparse, but not truly incremental:

- Anonymous guest memory includes all currently resident/faulted-in pages, including pages that were only read.
- A restored VM includes pages from the backing memory image that have merely been faulted in.
- Firecracker documents swap as a correctness limitation: swapped-out guest pages are not reported by `mincore` even though a snapshot may need them.

Consequently, a sandbox that changes only a small amount of memory between checkpoints may still write an OverlayBD layer approaching its resident working-set size. Frequent pause, snapshot, resume, and fork operations then pay unnecessary pause latency, storage I/O, object-storage/network transfer, and layer-compaction cost.

Firecracker PR #5274 explicitly distinguishes `mincore` snapshots from dirty-page tracked snapshots, and Firecracker issue #5768 describes the former as closer to sparse snapshots than differential snapshots:

- https://github.com/firecracker-microvm/firecracker/pull/5274
- https://github.com/firecracker-microvm/firecracker/issues/5768

I did not find an existing AgentENV issue or PR proposing true dirty-page page selection. AgentENV PR #92 adds optional compression, but compression reduces stored bytes rather than selecting only pages written during the current snapshot epoch, and it does not reduce the page set read from Firecracker.

### Use case

Long-running agent and RL sandboxes can have 1-8 GiB of guest memory and a large read-mostly working set. Between two checkpoints, the workload may modify only tens of MiB. The same environment may be checkpointed or forked many times during one trajectory.

For these workloads, each memory layer should contain approximately the pages actually written since tracking began, rather than the full resident working set. Reading a large restored working set should not by itself make the next memory layer large.

### Current behavior and workarounds

The default direct path is:

```text
pause VM
  -> create state-only Diff snapshot
  -> GET /vm/dirty-memory-ranges
  -> process_vm_readv
  -> compact_to
  -> overlaybd.commit
```

The legacy path creates a sparse `mem.bin` Diff snapshot and converts its data extents into `overlaybd.commit`.

Current mitigations are insufficient:

- `mincore` avoids copying never-resident anonymous pages, but cannot distinguish read-only resident pages from written pages.
- Compression (AgentENV PR #92) reduces artifact bytes but does not reduce the page set read from Firecracker, and it consumes additional CPU during pause.
- OverlayBD layer compaction controls some chain-growth cost but does not reduce the amount captured at each checkpoint.
- The native `mem.bin` path can reuse Firecracker's snapshot transaction, but adds an intermediate large-file write/read cycle before OverlayBD packaging.

Related AgentENV work:

- #92 compresses memory snapshot layers.
- #11 discusses transporting incremental memory-layer ancestry for VM migration, without changing page tracking.

### Desired behavior

- Add an explicit opt-in memory-snapshot tracking mode (for example `mincore` vs `dirty`) rather than silently changing the current default.
- In dirty mode, fresh VMs are configured with `track_dirty_pages = true` and file-backed restores are configured with `SnapshotLoadParams.track_dirty_pages = true`, so the selected page set comes from the KVM and Firecracker/VMM dirty bitmaps instead of `mincore`.
- Read-only resident pages are excluded from the captured set; only pages actually written are selected.
- Repeated captures of the same running sandbox remain complete relative to the launch-time parent snapshot: the dirty set may accumulate across captures, and this issue does not advance or reset the snapshot epoch (that is a separate issue).
- The runtime overhead of dirty-page logging is measured and documented before any default change is considered.

Acceptance criteria:

- Given a fresh VM with dirty mode enabled, when I write only a small set of pages and take a snapshot, then the memory layer contains those written pages and not the full resident set.
- Given a VM restored from a snapshot with dirty mode enabled, when I read a large portion of guest memory without writing it and take a snapshot, then the captured layer does not materially grow compared with `mincore` mode.
- Given two consecutive captures of the same running VM in dirty mode, when the snapshot data is inspected, then neither capture omits pages required for a complete restore relative to the launch parent (cumulative dirty set).
- A benchmark reports pause latency, selected page count, layer bytes, restore behavior, and steady-state runtime overhead for both `mincore` and dirty modes.

### Proposed approach

Start with an explicit opt-in mode rather than silently changing the current default. The exact config shape is open for discussion, for example:

```toml
[memory_snapshot]
tracking_mode = "mincore" # or "dirty"
```

The dirty mode would likely require:

1. Set `MachineConfiguration.track_dirty_pages = true` for fresh VMs.
2. Set `SnapshotLoadParams.track_dirty_pages = true` for file-backed restores.
3. Combine KVM and Firecracker/VMM dirty bitmaps, convert them to the existing destination `SegmentMapping` representation, and retain the current direct `process_vm_readv -> compact_to -> overlaybd.commit` data path.
4. Keep the running instance's dirty set cumulative relative to the original launch parent; do not reset or advance the dirty epoch in this issue so that repeated captures stay correct even before the epoch-advancement work lands.
5. Add comparative benchmarks for both modes before any default change.

Open questions for maintainers:

1. Is `mincore` deliberately preferred to avoid KVM dirty-logging overhead, or is it currently a compatibility/default choice?
2. In the custom Firecracker fork, does `GET /vm/dirty-memory-ranges` honor `track_dirty_pages` (KVM dirty log) and fall back to `mincore` when it is off?
3. Is an opt-in dirty mode with benchmarks an acceptable first PR before considering a default change?
4. Does AgentENV guarantee swap is disabled on runtime nodes, or should startup reject/warn when `mincore` mode is used with swap enabled?

### Compatibility and operational impact

API/config changes:

- Likely one memory-snapshot tracking-mode setting.
- No AgentENV public HTTP API change is expected.

Snapshot or storage format changes:

- None expected. Memory remains a bottom-up list of immutable OverlayBD layers.

New host/runtime requirements:

- KVM dirty logging must be supported for dirty mode.
- Runtime CPU/memory overhead of dirty-page logging must be benchmarked before changing the default.
- The `mincore` mode's swap precondition should be explicit.

Upgrade and rollback considerations:

- Start as opt-in so operators can return to `mincore` mode.
- Existing committed snapshots remain compatible.

### Alternatives considered

- **Keep `mincore` and rely on compression (AgentENV PR #92):** simple, but captures a read-resident superset and does not provide true write-set deltas; compression adds pause CPU.
- **Always enable dirty tracking with no config:** simplest, but imposes dirty-logging runtime overhead and portability constraints on all users; a toggle is more appropriate.
- **Compare resident pages with the parent:** avoids dirty-logging overhead but adds a large read/compare pass and complicates remote parent layers.
- **Track writes using userfaultfd or page protection:** substantially more complexity while Firecracker already supports KVM plus VMM dirty tracking.

### Non-goals

- Advancing the snapshot parent/epoch chain after each capture (separate issue).
- Changing the OverlayBD on-disk format.
- Changing rootfs or attached-drive snapshot behavior.
- Replacing compression or background download.
- Changing the default mode before benchmarks land.

### Contribution

I can implement it

### Pre-submission checklist

- [x] I searched existing issues and discussions and did not find a duplicate.
- [x] I described a concrete problem and use case, not only a proposed technology.
- [x] I understand that acceptance of the problem does not imply approval of a particular implementation.

Hygge-Gezelligheid:
## What

Add an opt-in `memory_snapshot.track_dirty_pages` configuration that enables Firecracker dirty-page tracking for direct memory snapshot capture.

- Pass the option when creating a fresh VM and when restoring a VM from a snapshot.
- Persist the option in `FirecrackerCommonConfig` so the capture behavior follows the sandbox that was launched with it.
- Require `direct_overlaybd=true` when dirty-page tracking is enabled.
- Reject dirty-page tracking in PVM mode because this combination has not been tested.
- Document the new configuration field, environment variable, defaults, and PVM limitation.

## Why

The current mincore-based capture path selects resident pages, including pages that were only read. For workloads with a large read-only resident set, this can copy substantially more memory than the guest actually modified.

Firecracker dirty-page tracking selects pages written after tracking begins. This provides the opt-in behavior requested by #119 while preserving the current behavior by default. The dirty bitmap is cumulative across captures, so a later snapshot can restore pages written both before and after an earlier capture.

## Related issue

Closes #119

Related to #120, but this PR does not implement per-snapshot dirty epochs.

## Scope and non-goals

Included:

- New `memory_snapshot.track_dirty_pages` boolean, defaulting to `false`.
- `AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES` environment-variable support.
- Configuration validation for the supported KVM + direct-overlaybd combination.
- Firecracker request propagation for fresh VM creation and snapshot restore.
- Backward-compatible persistence through a serde-defaulted common-config field.
- Unit tests for configuration parsing, mapping, and Firecracker Unix-socket request bodies.
- Configuration and PVM documentation updates.

Non-goals / intentionally excluded:

- PVM support before that combination is tested.
- Dirty-page tracking through the legacy `mem.bin` capture path.
- Per-capture dirty bitmap clearing or the snapshot-epoch model discussed in #120.
- Changes to snapshot manifests, OverlayBD layer formats, or public APIs.
- A claim of universal snapshot-performance improvement.

`MemorySnapshotConfig::track_dirty_pages` is mapped into `FirecrackerCommonConfig` and persisted with the sandbox. Fresh VM setup and snapshot restore both forward the persisted value to their corresponding Firecracker requests.

The direct capture decision uses either the current global `direct_overlaybd` setting or the persisted dirty-tracking setting. This prevents a sandbox launched with dirty tracking from falling back to the legacy capture path after a server restart or configuration change.

Configuration validation fails early for unsupported combinations:

- `track_dirty_pages=true` with `direct_overlaybd=false` is rejected because the legacy path cannot provide the required cumulative dirty-page behavior.
- `track_dirty_pages=true` in PVM mode is rejected with an explicit untested-combination error.

Existing configurations continue to use mincore behavior because the new setting defaults to `false`.

- Public API or generated protocol: N/A; no public API or generated protocol changes.
- Configuration or defaults: Adds an optional boolean and environment variable. The default is `false`, so existing deployments retain current behavior. Enabling it requires KVM and `direct_overlaybd=true`.
- Snapshot manifest, artifact layout, or storage format: Unchanged. The persisted Firecracker common config gains a serde-defaulted field, so existing persisted sandbox configurations remain readable.
- Upgrade and rollback: Upgrades preserve existing behavior until the option is enabled. Snapshot artifacts keep the existing format; disable the option before rolling back if the older deployment should not receive the new configuration field.
- Host requirements, permissions, ports, or dependencies: No new host dependencies, permissions, or ports. The option is limited to KVM because PVM has not been validated.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [ ] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
make fmt
# passed

make clippy
# passed

cargo test -p agentenv --lib -- --nocapture
# 692 passed, 0 failed, 4 ignored

git diff --check
# passed
```

Focused Firecracker socket tests assert that dirty tracking is enabled in both the fresh `/machine-config` request and the restore `/snapshot/load` request.

A 512 MiB anonymous guest-memory probe verified:

```text
read checksum after restore: passed
write/verify after restore: passed
s1_restores_A=true
s2_restores_A_and_B=true
DAMON-wait checksum: passed
```

Selected-byte comparison (selection/correctness diagnostic, not an end-to-end speedup claim):

| Workload | mincore selected bytes | dirty selected bytes |
| --- | ---: | ---: |
| quiet | 51,740,672 | 4,382,720 |
| read-only | 605,220,864 | 13,094,912 |
| write | 227,663,872 | 175,906,816 |
| cumulative S1 | 141,078,528 | 89,845,760 |
| cumulative S2 | 237,756,416 | 176,066,560 |

The read-only probe selected 97.8% fewer bytes with dirty tracking. In full-page-write ABBA runs, both modes selected about 619 MB, as expected. Sequential snapshot medians were 515.236 ms (mincore) and 517.641 ms (dirty), which did not show a clear regression. Random-write results were underpowered after excluding runs with swap I/O, so this PR does not claim a general performance improvement.

Skipped checks and reasons:

- `make test-unit`: not run; the focused `agentenv` library suite above was run instead.
- Repository Rust integration-test command: not run; the KVM correctness scenarios above were executed with the temporary experiment harness, which is not part of this PR.
- `make -C services test`: no `services/` changes.
- Generated clients/server regeneration: no schema or generated-code changes.

## Risks and reviewer notes

- Dirty-page tracking is cumulative: S2 includes pages dirtied before S1 as well as pages dirtied afterward. Per-snapshot epochs remain outside this PR and are related to #120.
- Performance benefit depends on workload. Read-heavy resident memory benefits in selected bytes; a workload that dirties nearly every page does not.
- PVM is intentionally rejected until it receives equivalent correctness validation.
- The legacy `mem.bin` path is intentionally rejected when dirty tracking is enabled.
- Most important review points: configuration validation in `src/cfg.rs`, persisted setting and Firecracker request propagation in `src/sandbox/firecracker/{config,instance}.rs`, and capture-path selection in `src/sandbox/firecracker/sandbox.rs`.

## Checklist

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -202,6 +202,12 @@ fill_concurrency = 4
 # Experimental: create memory overlaybd layers directly from Firecracker dirty
 # ranges via process_vm_readv, skipping the intermediate mem.bin file.
 direct_overlaybd = true
+# Experimental: enable Firecracker KVM dirty-page tracking for memory snapshots.
+# Requires direct_overlaybd = true. PVM is temporarily disabled because this
+# combination has not been tested.
+# Default false uses the mincore-based memory selection path.
+# Can also be set with AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES.
+track_dirty_pages = false
 compression_enabled = false
 compression_algorithm = "lz4"
 # Blocking threads used to compress memory-layer blocks. 1 = sequential;
--- a/config/oss_default.toml
+++ b/config/oss_default.toml
@@ -148,6 +148,12 @@ fill_concurrency = 4
 # Experimental: create memory overlaybd layers directly from Firecracker dirty
 # ranges via process_vm_readv, skipping the intermediate mem.bin file.
 direct_overlaybd = true
+# Experimental: enable Firecracker KVM dirty-page tracking for memory snapshots.
+# Requires direct_overlaybd = true. PVM is temporarily disabled because this
+# combination has not been tested.
+# Default false uses the mincore-based memory selection path.
+# Can also be set with AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES.
+track_dirty_pages = false
 compression_enabled = false
 compression_algorithm = "lz4"
 # Blocking threads used to compress memory-layer blocks. 1 = sequential;
--- a/docs/src/configuration/reference.md
+++ b/docs/src/configuration/reference.md
@@ -489,6 +489,7 @@ the default path on every startup.
 |-----|------|---------|-------------|
 | `overlaybd_global_config_path` | string | `"$AENV_HOME/overlaybd/mem-overlaybd-global.json"` | Path to the overlaybd global config used for the memory-snapshot ublk backend. Regenerated at startup (manual edits are overwritten); change only to relocate the generated file. |
 | `direct_overlaybd` | bool | `true` | Create memory overlaybd layers directly from Firecracker dirty memory ranges via `process_vm_readv`, skipping the intermediate `mem.bin` file. Set `AGENTENV_MEMORY_SNAPSHOT_DIRECT_OVERLAYBD=false` to force the legacy `mem.bin` conversion path. |
+| `track_dirty_pages` | bool | `false` | Enable Firecracker KVM dirty-page tracking for memory snapshots. When true, `direct_overlaybd` must also be `true`; PVM is temporarily disabled because this combination has not been tested. Set `AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES=true` to enable it. |
 | `compression_enabled` | bool | `false` | Enable compression for memory snapshot layers. When disabled, `compression_algorithm` is still parsed but has no effect. This setting affects only memory layers; the physical file name remains `overlaybd.commit`. |
 | `compression_algorithm` | string | `"lz4"` | Compression algorithm for memory snapshot layers. Valid values are only `lz4` and `zstd`. |
 
--- a/docs/src/deployment/pvm.md
+++ b/docs/src/deployment/pvm.md
@@ -223,6 +223,8 @@ virtualization_mode = "pvm"
 
 The environment variable takes precedence over the TOML value.
 
+Memory snapshot dirty-page tracking (`memory_snapshot.track_dirty_pages = true`) is temporarily disabled in PVM mode because this combination has not been tested.
+
 To build a PVM Docker image:
 
 ```bash
--- a/src/cfg.rs
+++ b/src/cfg.rs
@@ -415,6 +415,10 @@ pub struct MemorySnapshotConfig {
     pub overlaybd_global_config_path: PathBuf,
     #[config(env = "AGENTENV_MEMORY_SNAPSHOT_DIRECT_OVERLAYBD", default = true)]
     pub direct_overlaybd: bool,
+    /// Enable Firecracker KVM dirty-page tracking for memory snapshots.
+    /// Default: false, preserving the mincore-based path.
+    #[config(env = "AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES", default = false)]
+    pub track_dirty_pages: bool,
     #[config(default = false)]
     pub compression_enabled: bool,
     #[config(default = "lz4")]
@@ -880,6 +884,7 @@ impl AppConfig {
         if self.ublk.overlaybd.resize_timeout_secs == 0 {
             bail!("invalid ublk.overlaybd config: resize_timeout_secs must be > 0");
         }
+        self.validate_memory_snapshot_options()?;
         self.validate_memory_snapshot_background_download()?;
         self.validate_overlaybd_global_config_paths()?;
         self.validate_disk_rate_limit()?;
@@ -927,6 +932,25 @@ impl AppConfig {
                     i64::MAX
                 );
             }
+        }
+        Ok(())
+    }
+
+    fn validate_memory_snapshot_options(&self) -> Result<()> {
+        let memory = &self.memory_snapshot;
+        if !memory.track_dirty_pages {
+            return Ok(());
+        }
+        if self.virtualization_mode == VirtualizationMode::Pvm {
+            bail!(
+                "memory_snapshot.track_dirty_pages=true is disabled in PVM mode because this combination has not been tested"
+            );
+        }
+        if !memory.direct_overlaybd {
+            bail!(concat!(
+                "memory_snapshot.track_dirty_pages=true requires ",
+                "memory_snapshot.direct_overlaybd=true for cumulative dirty snapshots"
+            ));
         }
         Ok(())
     }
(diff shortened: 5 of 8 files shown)


huang-jl:
LGTM. Shall we mark this PR as ready for review (out of Draft)?

Hygge-Gezelligheid:
> LGTM. Shall we mark this PR as ready for review (out of Draft)?

Sure.

huang-jl (approved):

yingdi-shan (approved):

PR #145, opened 2026-08-09
fix(network): support iptables-legacy with CAP_NET_RAW

Park-Jiyeonn:
### Summary

iptables-legacy needs CAP_NET_RAW in addition to CAP_NET_ADMIN. Pass it through the runtime and scoped helpers so sandbox networking works on hosts using the legacy backend.


yingdi-shan:
We're not planning to support iptables-legacy. Its linear rule evaluation doesn't scale well once you have a large number of sandboxes.

PR #150, opened 2026-08-10
fix: sanitize singleton CPU config intersections

### LINKED ISSUE #137: Single-node CPU intersection bypasses KVM read-only leaf filtering and breaks sandbox creation
## Affected component

Gateway / scheduler, with impact on the Firecracker runtime and observability CPU-template reporting.

## AgentENV version

- Version or commit: `main` at `00ba6cb74f309a0cbd65f6ec415640e9e674a371` (`v0.1.1-22-g00ba6cb` according to `git describe`)
- Container image tag used by the deployment: a site-local convenience tag, not an AgentENV release version
- Local source modifications: none

## Host environment

- OS: Ubuntu 22.04
- Kernel: Linux 6.12.33 with PVM/KVM support
- Architecture: x86_64
- Bare metal, VM, or cloud: VM with PVM enabled
- ublk support: `ublk_drv` loaded and `/dev/ublk-control` available
- CPU: Intel Skylake-family virtual CPU, 16 vCPUs
- Memory: 30 GiB
- Storage/filesystem: ext4
- Kubernetes: single-node k3s cluster
- Firecracker: `v1.17.0-next.1` from the PVM dependency manifest

## Operation and configuration

```shell
# Relevant runtime configuration:
AENV_VIRTUALIZATION_MODE=pvm

# Observability and scheduler reporting are enabled. The Firecracker package
# contains an executable cpu-template-helper, so the node runs:
cpu-template-helper template dump -o /dev/stdout

# The scheduler discovers exactly one AgentENV node.

# After the node reports its CPU configuration and receives the scheduler result:
aenv start --cold -d <OCI_IMAGE>
```

## Steps to reproduce

1. Deploy Scheduler, Gateway, and one AgentENV node with Kubernetes discovery.
2. Leave observability scheduler reporting enabled and ensure the selected Firecracker dependency contains an executable `cpu-template-helper`.
3. Wait for the node log to report `received cluster cpu config intersection from scheduler`.
4. Create any cold-start sandbox.
5. Observe that Firecracker rejects the CPU configuration before boot.
6. As a control, make `cpu-template-helper` unavailable/non-executable, restart the scheduler and node, and repeat the same sandbox creation. It succeeds.

The failure reproduced consistently before disabling the helper. With the helper disabled, cold start, command execution, pause, resume, and memory-state preservation all succeeded.

## Expected behavior

A CPU configuration returned by the scheduler should be safe to apply to Firecracker. A single-node cluster should still run the same sanitization that removes KVM read-only/dynamic CPUID leaves.

## Actual behavior

Sandbox creation fails with HTTP 500. Firecracker returns:

```text
Start microvm error: System configuration error: Failed to create guest config: Template changes a CPUID entry not supported by KVM: Leaf: b, Subleaf: 1
```

## Root cause

`IntersectCpuConfigs` returns the only input unchanged when the cluster has one CPU report:

https://github.com/kvcache-ai/AgentENV/blob/00ba6cb74f309a0cbd65f6ec415640e9e674a371/services/scheduler/internal/cpu_template.go#L46-L52

That bypasses JSON validation and the existing KVM read-only leaf filter. The same file already classifies leaf `0xB` as read-only and removes it in the normal intersection path:

https://github.com/kvcache-ai/AgentENV/blob/00ba6cb74f309a0cbd65f6ec415640e9e674a371/services/scheduler/internal/cpu_template.go#L83-L85

https://github.com/kvcache-ai/AgentENV/blob/00ba6cb74f309a0cbd65f6ec415640e9e674a371/services/scheduler/internal/cpu_template.go#L120-L135

The current single-input unit test explicitly requires byte-for-byte passthrough and does not include a read-only CPUID leaf:

https://github.com/kvcache-ai/AgentENV/blob/00ba6cb74f309a0cbd65f6ec415640e9e674a371/services/scheduler/internal/cpu_template_test.go#L54-L63

Why leaf `0xB`, subleaf `1` appears in the dump:

1. Newer KVM versions intentionally omit `CPUID.(EAX=0BH, ECX=1)` from `KVM_GET_SUPPORTED_CPUID`.
2. Firecracker inserts that topology subleaf during its later CPUID normalization step.
3. `cpu-template-helper template dump` reads the already-configured vCPU and serializes every CPUID entry into custom-template JSON.
4. AgentENV sends the raw single-node dump back to Firecracker as a pre-boot custom CPU template. Template application happens against the KVM-supported CPUID set, before Firecracker's normalization step, so the injected topology entry is not present and is rejected.

Firecracker normalization source:

https://github.com/kvcache-ai/firecracker-next/blob/v1.17.0-next.1/src/vmm/src/cpu_config/x86_64/cpuid/normalize.rs#L265-L286

Helper dump conversion source:

https://github.com/kvcache-ai/firecracker-next/blob/v1.17.0-next.1/src/cpu-template-helper/src/template/dump/x86_64.rs#L18-L45

This was observed on PVM, but the control flow is not inherently PVM-specific. A recent standard KVM environment can be affected whenever its helper dump contains leaf `0xB`.

## Suggested fix

- Remove the `len(jsons) == 1` raw-return fast path, or pass a single input through a dedicated sanitizer before returning it.
- Ensure singleton input is parsed and removes `kvmReadOnlyLeafIDs`, while preserving all safe fields unchanged.
- Add a regression test where a single input contains leaf `0xB` and verify the output omits it.
- Consider validating singleton JSON as well; currently invalid JSON also bypasses parsing in the single-input path.
- Optionally run or expose `cpu-template-helper template verify` before publishing a generated cluster template.

## Frequency

Always, with the described single-node configuration.

## Regression information

- Last known good: unknown
- First known bad: present in the initial open-source implementation (`8f028b1369a4ce9a66d140174d0a646e72398517`)

## Additional context

The Firecracker documentation describes `template dump` as an entry point for creating a template and recommends stripping, reviewing, and verifying the result rather than replaying the raw dump directly:

https://github.com/firecracker-microvm/firecracker/blob/main/docs/cpu_templates/cpu-template-helper.md

AgentENV automates the dump/intersection/replay flow, so it needs to perform the required sanitization itself for every cluster size.

## Pre-submission checklist

- [x] I searched existing open and closed issues and did not find a duplicate.
- [x] I provided a minimal reproduction.
- [x] I removed credentials, tokens, private image references, and other sensitive information.
- [x] This is not a security vulnerability.

zlzGithub-0801:
## What

Fix the single-node startup failure when using the scheduler.

## Why

In a single-node setup with the scheduler enabled, CPU config intersection took the fast path and returned the config unchanged. This preserved CPUID leaf `0xB`, subleaf `0x1`, which caused Firecracker startup to fail.

## Related issue

Closes #137

## Scope and non-goals

Included:

- Update the scheduler CPU config intersection logic.

Non-goals:

- Changes to node logic.

Single-node CPU configs no longer take the fast path. They now use the same filtering and validation logic as multi-node configs.

- Public API or generated protocol: No changes.
- Configuration or defaults: No changes.
- Snapshot manifest, artifact layout, or storage format: No changes.
- Upgrade and rollback: No changes.
- Host requirements, permissions, ports, or dependencies: No changes.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] `make -C services test` (required when `services/` changes)
- [ ] Documentation updated

Commands and results:

Skipped checks and reasons:

Documentation updates and performance benchmarks were skipped because this change does not affect documentation or performance.

## Checklist

### DIFF
--- a/services/scheduler/internal/cpu_template.go
+++ b/services/scheduler/internal/cpu_template.go
@@ -41,14 +41,10 @@ type cpuidLeafKey struct {
 // IntersectCpuConfigs computes a conservative bitwise AND intersection of
 // Firecracker cpu_config_json strings. Only entries present in every input
 // config are retained; bitmap fields for shared entries are ANDed together.
-// Returns an empty string when jsons is empty; returns jsons[0] unchanged
-// when only one config is provided.
+// Returns an empty string when jsons is empty.
 func IntersectCpuConfigs(jsons []string) (string, error) {
 	if len(jsons) == 0 {
 		return "", nil
-	}
-	if len(jsons) == 1 {
-		return jsons[0], nil
 	}
 
 	configs := make([]cpuConfig, len(jsons))
--- a/services/scheduler/internal/cpu_template_test.go
+++ b/services/scheduler/internal/cpu_template_test.go
@@ -52,13 +52,15 @@ func TestIntersectCpuConfigs_Empty(t *testing.T) {
 }
 
 func TestIntersectCpuConfigs_Single(t *testing.T) {
-	input := buildConfig(nil, nil, nil)
+	safe := cpuidModifier{Leaf: "0x1", Subleaf: "0x0", Modifiers: []registerMod{{Register: "eax", Bitmap: bm32(1)}}}
+	input := buildConfig(nil, []cpuidModifier{safe, {Leaf: "0xb", Subleaf: "0x1", Modifiers: []registerMod{{Register: "eax", Bitmap: bm32(2)}}}}, nil)
+	want := buildConfig(nil, []cpuidModifier{safe}, nil)
 	out, err := IntersectCpuConfigs([]string{input})
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	if out != input {
-		t.Errorf("single config: want %q, got %q", input, out)
+	if out != want {
+		t.Errorf("single config: want %q, got %q", want, out)
 	}
 }
 


yingdi-shan (approved):
LGTM.

Issue #138, opened 2026-08-07
fix(p2p): sandbox veth churn kills the iroh endpoint ~1.3s after startup, disabling P2P for the process lifetime


dushulin:
## Summary

With `[p2p].enabled = true`, the iroh endpoint and the `iroh-blobs` store are torn
down **~1.3 s after node startup, before a single artifact is ever served**. On an
8-node cluster, 7 of 8 nodes died this way; the 8th survived only by winning a race.
Once dead, nothing ever restarts the transport, so P2P is silently unavailable for
the entire lifetime of the process and every overlaybd read falls back to the origin
registry.

The trigger is AgentENV's own warm-pool priming: the `veth-*` interfaces it creates
in the runtime pod's network namespace are seen by `netwatch` as a *major* link
change, which makes iroh rebind its UDP socket — and that rebind is not recoverable
when it fails.

## Environment

| | |
|---|---|
| AgentENV runtime | `v0.1.3` |
| iroh | `1.0.0-rc.0` (also reproduces by inspection on `1.0.3`) |
| netwatch | `0.19.1` (unmodified from crates.io) |
| Kernel | 6.8, cgroup v2 |
| Deployment | Kubernetes DaemonSet, pod has its **own** netns (`hostNetwork` unset) |
| Config | `[p2p] enabled=true, transport="iroh", listen_addr="0.0.0.0:0"`, `[snapshot] p2p_enabled=true`, `[ublk] download_enable=true` |
| Scale | 8 runtime nodes, all with identical config |

## Reproduction

Enable P2P with the default `listen_addr = "0.0.0.0:0"` and start the node. The
failing log lines are **debug-level only**, so an `info` log shows nothing at all —
this is why the failure went unnoticed. Use:

```
RUST_LOG=info,iroh=debug,iroh_blobs=debug,netwatch=debug,noq=debug
```

Observed on 7 of 8 nodes, marker counts byte-identical on every one of them
(`rebinding`=1, `Address already in use`=3, `socket closed`=2, `closing database`=1):

```
13.572  iroh::socket::transports::ip: binding addr=0.0.0.0:0
13.572  iroh::socket::transports::ip: successfully bound local_addr=0.0.0.0:51871
13.597  iroh::_events::direct_addrs: addrs={10.126.17.156:51871 Local}
13.597  iroh::endpoint: iroh endpoint bound iroh_version=1.0.0-rc.0
13.607  agentenv::p2p::iroh::transport: iroh artifact transport started
[...]

PR #152, opened 2026-08-11
fix(p2p): do not cache lookup timeouts as misses

WindSongShaoGe:
## Summary

- avoid inserting a negative descriptor-cache entry when a P2P lookup times out
- allow the next layer read to retry peer lookup instead of falling back to the origin for the full miss TTL
- add a regression test covering repeated reads after lookup timeouts

## Why

A lookup timeout means the artifact availability is unknown, not that the
artifact is absent. Caching the timeout as a miss causes subsequent reads for
that layer to fall back to the origin registry until the negative-cache entry
expires.

## Validation

- `make test`
- result: passed, exit code 0

Refs #138


yingdi-shan (approved):
LGTM. Thanks for your contribution.

PR #160, opened 2026-08-12
fix(api): accept allowInternetAccess as an alias so isolation requests can't silently fail open

### LINKED ISSUE #156: NewSandbox silently drops allowInternetAccess: sandbox created with open egress when isolation was requested
### Summary

`NewSandbox` declares `allow_internet_access` (snake_case) while the rest of the API — including the response model clients read back — uses `allowInternetAccess` (camelCase). A `POST /sandboxes` body that sets `allowInternetAccess: false` therefore has the field **silently dropped**, and the sandbox is created with the open base policy instead of the requested deny.

This is a fail-open outcome for a containment control, so flagging it as security-relevant rather than a cosmetic naming inconsistency.

### The inconsistency

`src/api/openapi.yml` uses both spellings across four schemas:

| Schema | Line | Spelling | Direction |
|---|---|---|---|
| `SandboxNetworkUpdateConfig` | 334 | `allow_internet_access` | request (update) |
| `SandboxDetail` | 446 | `allowInternetAccess` | **response** |
| `NewSandbox` | 541 | `allow_internet_access` | request (create) |
| `NewColdSandbox` | 588 | `allowInternetAccess` | request (cold create) |

The generated model faithfully reproduces the spec, so `src/api/generated/src/models.rs` carries `#[serde(rename = "allow_internet_access")]` on `NewSandbox`.

The two create paths disagree with each other, and the primary create path disagrees with the response.

### Why it fails open rather than erroring

From `src/api/impls/sandbox.rs`:

```rust
fn base_policy_from_allow_internet_access(value: Option<bool>) -> BaseSandboxNetworkPolicy {
    match value {
        Some(true)  => BaseSandboxNetworkPolicy::Allow,
        Some(false) => BaseSandboxNetworkPolicy::Deny,
        None        => BaseSandboxNetworkPolicy::Default,
    }
}
```

serde ignores the unknown camelCase key, so the field arrives as `None` → `Default`, not `Deny`. Nothing rejects the request and nothing warns.

The readback then makes it hard to notice: `allow_internet_access_from_base_policy(Default)` returns `Nullable::Null`, so `GET /sandboxes/{id}` reports `"allowInternetAccess": null` — which reads as "never set" rather than "your setting was discarded."

### Impact

A client that sends the camelCase spelling gets an internet-enabled sandbox after explicitly asking for an isolated one. That spelling is the natural choice, because it is what `SandboxDetail` returns, what `NewColdSandbox` accepts, and what an E2B-compatible client would send given E2B's camelCase conventions.

There is a second-order effect on domain egress: `allowOut` domain rules are only armed under a `Deny` base policy, so a create call combining `allowInternetAccess: false` with domain rules produces a sandbox with no egress restriction at all — neither the deny nor the allowlist.

### Reproduction

```bash
curl -sS -X POST http://127.0.0.1:8000/sandboxes \
  -H 'Content-Type: application/json' \
  -d '{"templateID":"ubuntu","allowInternetAccess":false}'

curl -sS http://127.0.0.1:8000/sandboxes/<id> | jq .allowInternetAccess
# → null   (expected: false)
```

From inside that sandbox, arbitrary outbound requests succeed.

Sending `allow_internet_access: false` instead behaves correctly, which is why the CLI path does not surface this — `aenv network` and the update model both use the snake_case spelling.

### Suggested fix

Two parts, and the second matters more than the first:

1. **Immediate:** rename the `NewSandbox` field to `allowInternetAccess` to match the spec's response and cold-create schemas, with `#[serde(alias = "allow_internet_access")]` so existing snake_case clients keep working. Same alias on `SandboxNetworkUpdateConfig` so both endpoints accept both spellings.
2. **Root cause:** settle the spec on one spelling. Right now the two create schemas disagree, so any regeneration reintroduces this.

Worth considering separately: should an unrecognised field in a network-policy request be rejected rather than ignored? A `deny_unknown_fields` posture on these specific models would turn this class of bug into a 4xx instead of an open sandbox.

### Offer

I have this fixed locally against v0.1.1 — the rename plus alias, alias on the update model, and regression tests covering camelCase and snake_case deserialization on both create and update, plus `Deny` base-policy propagation from a camelCase create body. Happy to open a PR against `main` if the approach above looks right to you, or to adjust it if you would rather fix the spec first and regenerate.

Verified present on `main` as of this writing.

sunkencity999:
Follows the direction you suggested in #156: `allow_internet_access` stays canonical on `NewSandbox` for E2B compatibility, and `allowInternetAccess` is accepted as a serde alias. Nothing is renamed and no existing client changes behaviour.

You were right that the field name is intentional — I checked the E2B spec you linked and it has the same split (`NewSandbox` takes snake_case, `SandboxDetail` returns camelCase), so AgentENV is mirroring the contract correctly.

### What this fixes

The residual problem is what happens when a caller sends the *other* spelling. serde drops the unknown key, so the field arrives as `None`; `base_policy_from_allow_internet_access` maps `None` to `BaseSandboxNetworkPolicy::Default` rather than `Deny`, and the sandbox is created with unrestricted egress after isolation was explicitly requested. Nothing rejects the request and nothing logs it.

`GET /sandboxes/{id}` then reports `allowInternetAccess: null`, which reads as "never set" rather than "your setting was discarded" — so it isn't obvious from the readback either.

Callers land on the wrong spelling because the contract is internally inconsistent: `allowInternetAccess` is what the response hands back and what `NewColdSandbox` accepts. The alias closes that trap without touching the canonical name.

There's a second-order effect worth noting: `allowOut` domain rules only arm under a `Deny` base policy, so a create combining `allowInternetAccess: false` with domain rules produced a sandbox with no egress restriction at all — neither the deny nor the allowlist.

### Changes

- `NewSandbox`: `#[serde(rename = "allow_internet_access", alias = "allowInternetAccess")]`
- `SandboxNetworkUpdateConfig`: same alias, so both endpoints behave alike
- Three regression tests: camelCase and snake_case on create, both spellings on update, and `Deny` base-policy propagation from a camelCase create body

Both model edits are in a generated file and carry comments so a regeneration doesn't silently revert them. If you'd rather express this in `openapi.yml` and regenerate — which is the more durable fix, and what you hinted at — I'm happy to close this in favour of that; the tests would still be worth keeping either way.

### Not addressed here

Whether an unrecognised field in a network-policy body should be rejected outright rather than ignored. `deny_unknown_fields` on these two models would turn this whole class of mistake into a 4xx instead of an open sandbox, and stays E2B-compatible since conforming clients only send known fields. It's a behaviour change, so it seemed wrong to bundle it — raised separately in #156 if you think it's worth pursuing.

Closes #156 if you take this approach; happy to adjust or withdraw.

### DIFF
--- a/src/api/generated/src/models.rs
+++ b/src/api/generated/src/models.rs
@@ -2874,7 +2874,13 @@ pub struct NewSandbox {
     pub secure: Option<bool>,
 
     /// Allow sandbox to access the internet. When set to false, it behaves the same as specifying denyOut to 0.0.0.0/0 in the network config.
-    #[serde(rename = "allow_internet_access")]
+    // `allow_internet_access` stays canonical, matching the E2B NewSandbox
+    // contract. The alias additionally accepts `allowInternetAccess` — the
+    // spelling SandboxDetail returns and NewColdSandbox takes — because an
+    // unrecognised key here is silently dropped, yielding
+    // BaseSandboxNetworkPolicy::Default (open) when the caller asked for Deny.
+    // Manual patch: preserve on regeneration.
+    #[serde(rename = "allow_internet_access", alias = "allowInternetAccess")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub allow_internet_access: Option<bool>,
 
@@ -5912,7 +5918,9 @@ pub struct SandboxNetworkUpdateConfig {
     pub deny_out: Option<Vec<String>>,
 
     /// Allow sandbox to access the internet. When set to false, it behaves the same as specifying denyOut to 0.0.0.0/0 in the network config.
-    #[serde(rename = "allow_internet_access")]
+    // Same alias as NewSandbox so both endpoints accept both spellings.
+    // Manual patch: preserve on regeneration.
+    #[serde(rename = "allow_internet_access", alias = "allowInternetAccess")]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub allow_internet_access: Option<bool>,
 }
--- a/src/api/impls/sandbox.rs
+++ b/src/api/impls/sandbox.rs
@@ -1498,4 +1498,42 @@ mod tests {
 
         assert_eq!(policy, SandboxNetworkPolicy::default());
     }
+
+    // Regression: the generated NewSandbox model renamed this field to
+    // "allow_internet_access" while the spec (and the aenv CLI) send
+    // "allowInternetAccess", so POST /sandboxes silently dropped it and
+    // sandboxes created with --no-internet came up with an open base policy.
+    #[test]
+    fn new_sandbox_accepts_camel_case_allow_internet_access() {
+        let body: models::NewSandbox = serde_json::from_str(
+            r#"{"templateID": "tpl", "allowInternetAccess": false,
+                "network": {"allowOut": ["198.51.100.7"]}}"#,
+        )
+        .expect("camelCase allowInternetAccess must deserialize");
+        assert_eq!(body.allow_internet_access, Some(false));
+
+        let policy = network_policy_from_create(body.allow_internet_access, body.network.as_ref())
+            .expect("policy from create");
+        assert_eq!(policy.base_policy, BaseSandboxNetworkPolicy::Deny);
+        assert_eq!(policy.egress.allowed_cidrs, ["198.51.100.7/32"]);
+    }
+
+    #[test]
+    fn new_sandbox_still_accepts_snake_case_allow_internet_access() {
+        let body: models::NewSandbox =
+            serde_json::from_str(r#"{"templateID": "tpl", "allow_internet_access": true}"#)
+                .expect("legacy snake_case spelling must keep deserializing");
+        assert_eq!(body.allow_internet_access, Some(true));
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


LSX-s-Software:
Thanks for preparing the patch.

The main issue for us is maintainability: `src/api/generated/` is fully owned by OpenAPI Generator. The aliases added here cannot currently be represented in `openapi.yml`, and the `rust-axum` generator will remove them whenever `cargo adev codegen server` is run. Comments in the generated file help explain the change, but cannot prevent that drift.

For that reason, we would prefer not to merge a manual modification to the generated models. We will keep `allow_internet_access` as the canonical `NewSandbox` wire field, consistent with the E2B OpenAPI contract and SDK request serialization.

A good path forward may be to contribute field-alias support to the upstream [`rust-axum` generator](https://github.com/OpenAPITools/openapi-generator). That would let the alias live in the OpenAPI source of truth and be reproduced by every code-generation run. If upstream accepts such support, we would be happy to reconsider using it here.

The same applies to strict unknown-field handling: upstream support for generating `#[serde(deny_unknown_fields)]` from closed object schemas would provide a much more maintainable foundation than patching individual generated models.

Given the current generator limitations, we will close this PR for now. Thanks again for investigating the behavior and putting together the tests.

PR #163, opened 2026-08-12
experiment(cli): evaluate clap dynamic completion

NickNYU:
## Summary

This is one of two parallel experiments for #37. It evaluates `clap_complete` `unstable-dynamic` for dynamic sandbox-ID completion while preserving the existing static `aenv completion bash|zsh|fish` command.

## What it adds

- `CompleteEnv` shell adapter for Bash, Zsh, and Fish.
- Dynamic candidates backed by the existing sandbox list API.
- State-aware candidates for running, paused, and active sandbox IDs.
- Short completion-specific network timeouts and silent failure behavior.
- Focused tests and a design note.

## Trade-off

This keeps the shell protocol small and shell-portable, but depends on `clap_complete` unstable-dynamic and its generated shell adapter contract. The companion PR proposes an AgentENV-owned `aenv __complete` protocol instead. These are alternatives; please review and choose one direction rather than merging both.

## Verification

- `cargo fmt --all`
- `cargo check -p aenv --bin aenv`
- `cargo test -p aenv --bin aenv` (56 passed; one pre-existing macOS `/var` symlink test failure in download tests)


crates/aenv/src/commands/completion.rs
```diff
@@ -37,6 +44,70 @@ pub fn run(args: Args) -> Result<()> {
...
+    sandbox_candidates(|state| state == Some("running"))
+}
+
+pub fn paused_sandbox_candidates() -> Vec<CompletionCandidate> {
+    sandbox_candidates(|state| state == Some("paused"))
+}
+
+pub fn active_sandbox_candidates() -> Vec<CompletionCandidate> {
+    sandbox_candidates(|state| matches!(state, Some("running") | Some("paused")))
+}
```
LSX-s-Software, line 57: Since a sandbox can only be in one of two states—running or paused—this filter seems a bit redundant.

crates/aenv/src/commands/completion.rs
```diff
@@ -3,8 +3,15 @@ use anyhow::Result;
...
 use clap::ValueEnum;
+use clap_complete::engine::{ArgValueCandidates, CompletionCandidate};
 use clap_complete::Shell as ClapShell;
 use std::io::Write;
+use std::time::Duration;
+
+use crate::client::sandboxes::ListedSandbox;
+
+const DYNAMIC_CONNECT_TIMEOUT: Duration = Duration::from_millis(250);
+const DYNAMIC_REQUEST_TIMEOUT: Duration = Duration::from_millis(500);
```
LSX-s-Software: I think the timeout is a bit too short for remote users. Perhaps 500ms for connect and 1s for request is more appropriate.

crates/aenv/src/commands/snapshot.rs
```diff
@@ -30,7 +31,10 @@ enum Sub {
     #[command(visible_alias = "ls")]
     List {
         /// Filter snapshots by source sandbox ID
-        #[arg(long = "sandbox-id")]
+        #[arg(
+            long = "sandbox-id",
+            add = crate::commands::completion::add_active_sandbox_candidates()
```
LSX-s-Software: Please remove this completion candidate since the lifetime of the sandbox and its snapshots are independent.

LSX-s-Software (changes requested):
Thanks for contributing. I think this approach is more elegant and maintainable. However, there are some issues that should be fixed. In addition, please fix the `clippy` error and remove the docs generated by superpowers.

crates/aenv/src/commands/completion.rs
```diff
@@ -37,6 +44,68 @@ pub fn run(args: Args) -> Result<()> {
     write_completion(args.shell, &mut std::io::stdout().lock())
 }
 
+pub fn running_sandbox_candidates() -> Vec<CompletionCandidate> {
+    sandbox_candidates(|state| state == Some("running"))
+}
```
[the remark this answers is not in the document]
LSX-s-Software: @NickNYU Please check this (but ignore the suggestion). This is a real bug. Perhaps you should make `write_completion` write the dynamic engine's registration script.

NickNYU:
@LSX-s-Software The registration-script bug is fixed in 044f945.

**Root cause**: `write_completion` emitted a static `clap_complete::generate` script, which bakes the command tree into the script but never calls back into the binary — so the `ArgValueCandidates` providers were unreachable through the normal install path.

**Fix**: `write_completion` now writes the dynamic engine's registration script via `EnvCompleter::write_registration`, so `aenv completion <shell>` output hooks the shell to call `COMPLETE=<shell> aenv -- ...` at completion time. Verified byte-identical to what `COMPLETE=<shell> aenv` emits (when invoked via PATH).

One design note: the script invokes `aenv` by name (PATH resolution) rather than an absolute path, matching `CompleteEnv`'s default behavior for PATH-invoked binaries and keeping cached scripts working across in-place upgrades.

Verification:
- `cargo fmt --check`, `cargo clippy -p aenv --all-targets -- -D warnings`
- 12/12 completion tests pass (script assertions updated to check the `COMPLETE=` callback for bash/zsh/fish)
- Generated scripts for all three shells confirmed to contain the callback

Ready for another look.

LSX-s-Software (approved):

NickNYU:
Update on the latest CI run and review comment:

**Re: OpenCodeReview caching suggestion** — I'd prefer not to add a TTL/debounce cache here. Issue #37 explicitly lists "persistent cross-process completion caches" as a non-goal, and each Tab press spawns a fresh `aenv` process, so in-process caching can't help; a temp-file cache would be exactly the excluded non-goal. The short connect/request timeouts (500ms/1s, per the earlier review feedback) plus silent-failure are the agreed mitigation for a slow or unavailable API.

**Re: CI failures** — both look unrelated to this PR:

- `clippy`: the run picked up Rust 1.98.0 (released 2026-08-18), whose new `chunks_exact_to_as_chunks` lint fires on `storage/overlaybd/src/compression/zfile.rs:1320` and `storage/overlaybd/src/prefetch.rs:254`. That code is byte-identical on `main`; the last green main CI (2026-08-20) still ran 1.97.1. This is toolchain drift affecting the whole repo.
- `agentenv-tests`: `fc::microvm_can_access_internet` fails with `iptables-restore v1.8.7 (nf_tables): unknown option "--to"` — a CI runner environment issue; this PR only touches the `aenv` CLI crate.

Happy to either (a) wait for a main-side fix and rebase, or (b) include the two-line lint fix here (`chunks_exact(4)` → `as_chunks::<4>().0`) to get CI green — whichever you prefer, @LSX-s-Software.

LSX-s-Software:
It's OK to ignore the CI failures because they are unrelated to this PR and I will fix them in another PR. Thanks for your patience and all the work on this! This PR is already merged.

PR #176, opened 2026-08-17
Support configurable S3 addressing style

guozy18:
## Summary

Add an optional OSS/S3 bucket addressing-style override and propagate it through both snapshot data paths.

## Motivation

The existing OSS backend chooses the addressing style from the endpoint:

- Alibaba OSS endpoints and bucket-in-endpoint hosts use virtual-host style.
- Other endpoints fall back to path style.

Some S3-compatible providers require or prefer virtual-host addressing but cannot be identified reliably by the endpoint heuristic. Without an explicit override, those providers may generate invalid requests.

This change allows users to explicitly select the required addressing style.

## Changes

- Add the optional `backend.oss.addressing_style` configuration:
  - `virtual`
  - `path`
  - unset for the existing endpoint-based auto-detection
- Propagate the setting to the snapshot OSS client.
- Propagate the setting to the generated OverlayBD runtime configuration as `ossConfig.defaultAddressingStyle`.
- Apply the setting to OverlayBD remote managed-layer reads as well as snapshot repository operations and image export.
- Reject unsupported addressing-style values.
- Document the configuration and preserve the existing default behavior.
- Add a wire-level regression test that verifies the actual HTTP method, request path, and `Host` header.
- Keep the implementation focused by removing tests that only duplicated internal enum mappings or parser implementation details.

## Configuration example

```toml
[backend.oss]
endpoint = "https://t3.storage.dev"
bucket = "agentenv-snapshots"
region = "auto"
addressing_style = "virtual"
```

When `addressing_style` is omitted, the existing endpoint-based detection remains unchanged.

## Backward compatibility

Existing configurations do not require any migration.

The default behavior is preserved:

- recognized Alibaba/bucket-host endpoints continue to use virtual-host addressing;
- other endpoints continue to use path-style addressing unless explicitly overridden.

This change does not modify Prometheus metrics or the `/metrics` endpoints.

## Validation

The following checks pass locally:

- `git diff --check`
- `cargo test -p object-store-operator`
- `cargo clippy -p object-store-operator --all-targets -- -D warnings`

The full AgentENV and OverlayBD test suites require a Linux environment. They could not be run locally on macOS because the workspace's `io-uring` dependency uses Linux-only libc symbols. Linux CI is required for the full workspace verification.

## Relationship to #23

This PR supersedes #23.

The original requirement and initial implementation were proposed by @davidmyriel in #23. Thank you for that work and for the original motivation behind the change.

This PR carries the same behavior forward on top of the current `main` branch, with a refreshed implementation, focused regression tests, and updated documentation. The current implementation was reorganized to fit the latest codebase while preserving the original user-facing intent.

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -237,6 +237,9 @@ p2p_enabled = true
 # endpoint = "https://oss-cn-beijing-internal.aliyuncs.com"
 # bucket = "agentenv-snapshots"
 # region = "cn-beijing"
+# Bucket addressing style: "virtual" or "path". Leave unset to use
+# endpoint-based auto-detection.
+# addressing_style = "virtual" # or "path"
 # prefix = "snapshots/"
 # credential_process = "/usr/local/bin/get-oss-credentials"
 # access_key_id = "..."
--- a/crates/e2e-tests/tests/snapshot_oss_e2e_test.rs
+++ b/crates/e2e-tests/tests/snapshot_oss_e2e_test.rs
@@ -123,6 +123,7 @@ fn test_oss_config(fixture: &MinioFixture, prefix: &str) -> OssBackendConfig {
         access_key_secret: Some(MINIO_PASS.to_string()),
         security_token: None,
         region: Some(fixture.region.clone()),
+        addressing_style: None,
         cache_max_size_gb: Some(1),
     }
 }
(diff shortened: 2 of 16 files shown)


huajq (approved):
lgmt

PR #199, opened 2026-08-20
Recover snapshot cache after canceled restores

nikhilsimha:
## What changed

When the task leading a snapshot artifact materialization is canceled, the next waiter now removes that abandoned in-flight entry and retries. The cleanup compares the watch channel before removal so a concurrent waiter cannot remove a newer fetch.

A regression test cancels the materialization leader, then verifies a second caller materializes and reads the artifact.

## Why

A client disconnect can cancel the leader after the cache records the fetch but before it publishes a result. The sender is dropped, while the receiver remains in the in-flight map. Every later restore of that snapshot then fails with `ended without a result` until the AgentENV server restarts.

This change makes canceled restores self-healing and prevents a transient disconnect from permanently blocking sandbox recovery on that node.

## Validation

- `cargo test snapshot::artifact_cache::tests` — 7 passed
- `cargo clippy --locked --all-targets -- -D warnings` — passed
- `cargo test --locked` — 741 passed, 4 ignored; 5 unrelated ublk tests could not connect to the local ublk daemon (`channel closed`)

### DIFF
--- a/src/snapshot/artifact_cache.rs
+++ b/src/snapshot/artifact_cache.rs
@@ -160,9 +160,13 @@ impl LocalArtifactCache {
 
             let sender = {
                 let mut inflight = self.inflight.lock().await;
-                if let Some(existing) = inflight.get(key).cloned() {
+                if let Some(mut existing) = inflight.get(key).cloned() {
                     drop(inflight);
-                    wait_for_inflight_result(existing.receiver, key).await?;
+                    if wait_for_inflight_result(&mut existing.receiver).await?
+                        == InflightWait::Abandoned
+                    {
+                        self.clear_abandoned_inflight(key, &existing.receiver).await;
+                    }
                     continue;
                 }
 
@@ -243,8 +247,12 @@ impl LocalArtifactCache {
                 let inflight = self.inflight.lock().await;
                 inflight.get(key).cloned()
             };
-            if let Some(existing) = existing {
-                wait_for_inflight_result(existing.receiver, key).await?;
+            if let Some(mut existing) = existing {
+                if wait_for_inflight_result(&mut existing.receiver).await?
+                    == InflightWait::Abandoned
+                {
+                    self.clear_abandoned_inflight(key, &existing.receiver).await;
+                }
                 continue;
             }
             break;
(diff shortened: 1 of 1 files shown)


nikhilsimha:
Production validation on term.so:

- Deployed commit `e588f7e` to both current AgentENV nodes.
- `TestAgentENVFleetRuntimeLive` passed against the real gateway: create → write → publish/remove → restore under a new runtime ID → read → terminal ready marker.
- The user snapshot that originally failed (`01a01e20-5b47-7fb3-8344-31a69a160c65`) restored successfully; the diagnostic runtime was removed without reading workspace contents.
- Both node binaries have a rollback copy and the shared API key is mounted with `0640 root:aenv`.

src/snapshot/artifact_cache.rs
```diff
@@ -320,6 +328,20 @@ impl LocalArtifactCache {
         inflight.remove(key);
```
guozy18, line 328: ### Issue 2: Cancellation window between `send` and `remove`

`send` → `lock().await` → `remove` leaves a cancellation window: if the finisher is canceled at the `.await`, the map keeps a dead entry that already carries `Some(Success/Failed)`. Stale `Failed` makes every later request replay the same old error forever; stale `Success` spins `ensure_cached_at` in a miss → Finished → continue loop once the index entry is evicted.

Moving `send` inside the lock fixes it — no `.await` between `send` and `remove`:

```rust
let mut inflight = self.inflight.lock().await;
let _ = sender.send(Some(result));
inflight.remove(key);
```

If the task is canceled before acquiring the lock, the sender is dropped without a result, which your Abandoned path already handles.

src/snapshot/artifact_cache.rs
guozy18, line 185: ### Issue 1: Partial files are treated as successful fetches

`fetch` writes directly to the final path, but whether the file is usable is decided only by `try_exists`. These two don't compose safely:

1. The leader is canceled mid-write (or fails with `Err` after writing partial data) → the final path exists, but the content is incomplete.
2. With this PR, the next waiter clears the abandoned inflight entry and retries.
3. On retry, `try_exists(local_path)` returns true → `pin_local_file` registers the partial file in the index and returns a handle — without ever re-running `fetch`.

This turns an explicit failure into a silent success: e.g. resuming a sandbox from a truncated `vm_state.bin`, which will fail much later and be far harder to diagnose than a restore error. This is reachable in production — the OSS resolver's P2P export writes chunks straight to the target path.

guozy18 (changes requested):
This PR fixes a real issue: when a materialization leader is canceled, it leaves a stale entry in the inflight map that permanently poisons the snapshot key, causing every later restore to fail until a restart. The Abandoned handling and the `same_channel` ABA protection are the right approach. That said, I found two correctness issues that I'd like to see fixed before approving — I've left inline comments at the corresponding locations.

> **Note:** the first one roots in pre-existing code, but it only becomes reachable because of this PR's new retry path — previously the stale entry failed everything loudly, so a partial file could never be registered as a valid cache entry. Since this PR turns that loud failure into a potential silent success, I think it needs to be addressed here.

PR #217, opened 2026-08-25
fix(overlaybd): preallocate cache blocks to avoid SIGBUS when disk is…

huang-jl:
## What

Reserve the block's byte range with fallocate(FALLOC_FL_KEEP_SIZE) before handing out mmap write access (write_block and mmap_mut_for_block). Disk exhaustion now surfaces as a plain ENOSPC refill error, and reads fall back to the source exactly like the existing cache-capacity-full path. Filesystems without fallocate support (EOPNOTSUPP/ENOSYS) keep the previous behavior.

## Why

The full-file cache writes refilled blocks through a shared mmap over a sparse data file. Writing into an unallocated hole makes the kernel allocate filesystem blocks at page-fault time; when the local disk is full that allocation fails and the process is killed with SIGBUS, taking down the whole ublk-daemon.

Before write into mmap region of cache files, an extra `fallocate` syscall will be invoked.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [ ] Documentation updated

## Checklist

- [ ] The PR contains one coherent change and no unrelated formatting or refactoring.
- [ ] New behavior is covered by tests, or I explained why testing is impractical.
- [ ] Logs and examples contain no credentials, tokens, or private registry information.
- [ ] I did not manually edit generated code without updating its source and regenerating it.

### DIFF
--- a/storage/overlaybd/src/backend/cache/full_file_cache/cache_entry.rs
+++ b/storage/overlaybd/src/backend/cache/full_file_cache/cache_entry.rs
@@ -78,6 +78,64 @@ pub(crate) enum AcquireRefillResult {
 }
 
 // ---------------------------------------------------------------------------
+// Disk range reservation
+// ---------------------------------------------------------------------------
+
+/// Outcome of preallocating a byte range of the sparse data file.
+pub(crate) enum RangeAllocation {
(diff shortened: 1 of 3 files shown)


huajq (approved):
lgmt

PR #226, opened 2026-08-27
feat(image): publish converted OCI layers over P2P

LSX-s-Software:
## What

Publish converted standard OCI image layers through the existing P2P transport and reuse them across nodes. Track P2P artifact ownership in the image-cache metadata so eviction can remove remote references together with local commit files.

## Why

Standard OCI layers currently need to be downloaded and converted independently on each node. Sharing completed OverlayBD conversions avoids repeated registry downloads and conversion work while preserving the existing local-cache and registry fallback paths.

## Related issue

N/A. No issue or prior design discussion is referenced by the commits in this branch.

## Scope and non-goals

This PR includes:

- A stable `oci-layer/v1/{context-hash}` P2P key derived from `LayerConversionKey`.
- Reference-mode publication of durable converted OverlayBD commits with protocol, digest, and size metadata.
- P2P lookup, download, validation, local persistence, and conversion-index reuse.
- Persistent P2P-key ownership on image-cache hard-commit records and unpublish-before-delete during GC.
- The `iroh` 1.0.3 and `iroh-blobs` 0.103.0 dependency update.

This PR does not change overlaybd-native image handling, snapshot artifact publication, public API routes, or introduce new configuration knobs.

The server initializes the image cache with the same process-wide P2P transport used by the other P2P consumers. For a standard OCI layer, the conversion context is serialized and hashed into the catalog key. Once the converted commit is durable locally, the image cache records ownership and publishes the file by reference.

On a cache-index miss or stale entry, the image cache first attempts the corresponding P2P lookup. Fetched bytes must match the advertised digest and size and contain a sealed OverlayBD layer with the expected UUID before they are persisted and indexed locally. Lookup, fetch, metadata, digest, or structural validation failures are treated as acceleration misses and fall back to registry download and conversion; invalid fetched artifacts are unpublished when possible.

Image-cache GC carries the recorded P2P keys through its collectability checks and unpublishes them before removing the local commit and metadata. Metadata updates are serialized so ownership is preserved across concurrent record/reconcile operations.

- Public API or generated protocol: N/A; no public HTTP or generated protocol changes.
- Configuration or defaults: N/A; uses the existing P2P transport/configuration and adds no user-facing setting.
- Snapshot manifest, artifact layout, or storage format: No snapshot changes. Image-cache hard-commit metadata gains an additive optional `p2p_keys` field, defaulting to empty for existing records.
- Upgrade and rollback: The branch updates `iroh` to 1.0.3 and `iroh-blobs` to 0.103.0. Nodes without usable P2P publication or lookup continue through the existing local/registry paths; cache metadata remains readable when no P2P keys are present.
- Host requirements, permissions, ports, or dependencies: No new host requirement or port. Existing P2P and registry access prerequisites still apply.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
The author reports that the relevant tests were run successfully before PR creation. Tests were not rerun during PR preparation, per request; exact command output is not available in this shell.
```

Skipped checks and reasons:

```text
Formatting, clippy, and test commands were not rerun during PR creation because the author had already run the relevant validation. Services tests and code generation are not applicable because this branch does not change services or generated clients/server code. No benchmark was requested or reported.
```

## Risks and reviewer notes

- P2P publication is intentionally best-effort for image acceleration; registry conversion remains the correctness fallback.
- Remote bytes are validated against metadata, file digest, sealed OverlayBD structure, and the conversion-context UUID before reuse.
- Ownership is recorded before publication to avoid leaving a catalog reference without GC metadata after a crash; a failed publication may leave an unserved ownership record that is harmless and will be cleaned with the cache object.
- Review the conversion-key stability and the ordering of unpublish versus local deletion in `src/image/oci_image.rs` and `src/image/cache/service.rs`.

## Checklist


src/image/cache/service.rs
```diff
@@ -241,13 +251,63 @@ impl ImageCacheService {
...
+    /// P2P catalog. A missing transport is a normal no-op; callers can keep the
+    /// local conversion path usable when P2P is disabled.
+    async fn publish_converted_layer(
+        &self,
+        key: &LayerConversionKey,
+        layer: &LocalLayer,
+    ) -> Result<bool> {
+        let Some(transport) = self.p2p_transport.get().cloned() else {
+            return Ok(false);
+        };
```
guozy18, line 288: This early return is dead code when P2P is disabled, which is the default. Startup calls `initialize_image_cache_p2p_transport` no matter what, and in the disabled case that installs a `DisabledP2pTransport` whose `publish` just returns `Ok(())`. So the guard never triggers.

That has a couple of annoying consequences. On a default deployment, every local conversion records the key via `add_hard_commit_p2p_key` (the next line) even though nothing will ever be published, so the metadata slowly fills with ownership entries for commits nobody owns. Worse, cleanup of those keys is gated on the transport being initialized at all — `delete_collectable_hard_commit` refuses to delete them from a process that never initialized. So the entries are written unconditionally but can only be removed under a condition that has nothing to do with whether they exist.

I'd suggest checking whether the transport is actually enabled instead of whether it exists — e.g. return `Ok(false)` early when `local_endpoint().is_none()`.

src/image/cache/service.rs
```diff
@@ -1188,7 +1382,7 @@ impl ImageConversion for ImageCacheOperationHold {
                 path = %index_path.display(),
                 "converted user image layer index is missing from cache"
             );
-            return Ok(None);
+            return self.lookup_remote_converted_layer(key).await;
```
guozy18, line 1395: This path does the same failed lookup twice. `cached_standard_oci_lowers` returns `None` on the first miss, then `convert_standard_oci_layers_pipeline_inner` starts over from layer 0 and calls `lookup_converted_layer` for that same layer again. `ImageCacheOperationHold` doesn't remember per-operation results, so the first-missing layer's P2P lookup and fetch run twice in full inside a single resolve.

That gets expensive with a stale catalog entry. If the provider is dead, the fetch sits on the default 30s timeout, and every cold resolve of that image pays the stall twice before falling back to the registry. One bad catalog entry doubles the worst-case latency for everyone.

I'd cache failed P2P results by key in the hold, scoped to the operation's lifetime, so the second lookup gives up immediately instead of retrying the fetch.

guozy18 (changes requested):

guozy18 (approved):
LGTM.

PR #229, opened 2026-08-29
api: accept the E2B SDK's deprecated alias as the v3 template name

nblintao:
## What

`POST /v3/templates` accepts the template name as `name` or as the deprecated `alias`, instead of `name` only. The OpenAPI schema and the generated model gain the `alias` field, copied from E2B's schema; the handler resolves the name through one helper with a unit test.

## Why

E2B's current schema (`spec/openapi.yml` in e2b-dev/infra) defines `TemplateBuildRequestV3` as `name` + `tags` + `alias`, the last marked deprecated ("use name instead"), and E2B's own API still accepts it. AgentENV implements `name` and `tags` but not `alias`.

The e2b Python SDK switched from sending `alias` to sending `name` somewhere after 2.10.2 — the last release that supports Python 3.9. Current releases (2.46.0) work against AgentENV as-is; a client on Python 3.9, or on any SDK ≤ 2.10, gets `400 template name must be provided` and cannot build a template at all, while everything after that (create from the alias, exec, host routing) works. Accepting the deprecated field, as E2B does, closes that gap without changing what `name` means.

## Related issue

None filed. Reproducer: `Template.build` from e2b 2.10.2 (Python 3.9) against v0.1.3.

## Scope and non-goals

Included: schema, the generated model field (hand-added in the generator's style, since the model is checked in), name resolution in `v3_templates_post`. Not included: template tags (still rejected by the existing check), `teamID`, and any change to the v2 build-start request.

`name` wins over `alias`; an empty string counts as absent. No other data flow changes.

- Public API or generated protocol: one optional, deprecated field added to `TemplateBuildRequestV3`, matching E2B's schema; existing clients sending `name` are unaffected.
- Upgrade and rollback: stateless; rolling back only re-breaks the old SDKs.

## Verification

`make fmt`, `cargo clippy -p agentenv --lib -- -D warnings`, and the new unit test pass. Against a v0.1.3 cluster, e2b 2.10.2's request fails today and succeeds once the body also carries `name` (tested by patching the request client-side); e2b 2.46.0 builds and runs templates against the unpatched server.

### DIFF
--- a/src/api/generated/src/models.rs
+++ b/src/api/generated/src/models.rs
@@ -7947,6 +7947,12 @@ impl std::convert::TryFrom<HeaderValue> for header::IntoHeaderValue<TemplateBuil
 #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize, validator::Validate)]
 #[cfg_attr(feature = "conversion", derive(frunk::LabelledGeneric))]
 pub struct TemplateBuildRequestV3 {
+    /// Alias of the template. Deprecated, use name instead.
+    #[serde(rename = "alias")]
+    #[validate(length(max = 128), custom(function = "check_xss_string"))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub alias: Option<String>,
+
     /// Name of the template. Can include a tag with colon separator (e.g. \"my-template\" or \"my-template:v1\"). If tag is included, it will be treated as if the tag was provided in the tags array.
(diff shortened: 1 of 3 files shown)


nblintao:
Closing: the gap this addressed only exists for e2b ≤ 2.10.2 (the last release for Python 3.9). Current releases send `name` and work against the server as-is, so the change is not worth carrying. Leaving the branch in case anyone on an old SDK needs it.

PR #232, opened 2026-08-30
feat: add single-active scheduler HA

ethe:
## Summary

- add Kubernetes Lease leader election for full Scheduler replicas
- require Redis-backed sandbox bindings and Kubernetes discovery when leader election is enabled
- keep followers out of the serving gRPC endpoints and reject non-health RPCs
- require fresh READY heartbeats before a newly elected Scheduler places new work
- preserve existing-sandbox routing through independent query-only replicas

## Why

Running multiple full Scheduler replicas behind one Service currently creates independent in-memory node heartbeat/resource views, round-robin cursors, and artifact indexes. Even with Redis-backed bindings, active-active scheduling can use stale or incomplete capacity state. This change makes the full Scheduler single-active while retaining query-only data-plane availability.

## Validation

- `make -C services test`
- `go test ./...` from `services/`
- `go vet ./...` from `services/`

## Deployment contract

Leader-election deployments need a Service whose readiness probe uses gRPC health, Lease RBAC for the full Scheduler service account, a unique Pod identity through `SCHEDULER_LEADER_ELECTION_IDENTITY`, Redis-backed bindings, and separate query-only replicas for uninterrupted `LookupNode`. Observed node/resource state and the P2P artifact index remain intentionally ephemeral and are rebuilt after failover.


ethe:
Closing because this change is intended for the tonbo-io/AgentENV fork, not the upstream repository. The fork-local PR is tonbo-io/AgentENV#6.

PR #244, opened 2026-09-03
fix(overlaybd): restore premerged index prune offload lost in a rebase

YXalix:
## Problem

`uvm-ublk-daemon` RSS keeps climbing (observed close to 1 TiB) once the premerged-index cache directory accumulates hundreds of thousands of artifacts, and it stays high long after every sandbox has exited. The directory itself is only ~4 GiB.

Root cause: the prune offload from #179 was **silently reverted by #123**. The auth branch in #123 was authored before #179 merged; when it was rebased onto main before merging, the rebase conflict in `storage/overlaybd/src/lsmt/file/helper.rs` was resolved with the branch's stale copy. The revert is visible in #123's own diff as `helper.rs +24/-156` and `tests.rs -54` — an exact mirror of #179 — hidden inside a 24-file auth PR. Current `main` therefore runs the pre-#179 code.

## Why the old code leaks memory

The per-key lock only deduplicates writes for the *same* cache key. Every pause seals a new upper layer (new UUID), so every layer stack gets a *new* key — hundreds of thousands of artifacts means hundreds of thousands of distinct keys, each spawning its own detached write task with no global bound:

1. Each spawned task captures the whole merged `Arc<ReadOnlyIndex>` (~24 B/mapping; page-granular memory-snapshot stacks can reach hundreds of MB) and pins it until the task actually finishes.
2. Each task then runs a full-directory prune via async `tokio::fs`: one blocking-pool round-trip **per directory entry**. With N tasks and N entries that is O(N²) blocking work, which the 512-thread blocking pool can never drain.
3. Tasks are detached, so sandbox teardown cancels nothing; the backlog drains at a crawl after all sandboxes exit, so RSS stays elevated. The directory overshooting the `max_dir_bytes` cap (≤1 GiB configured, 4 GiB observed) confirms prune was no longer keeping up.

## Fix

Restore the #179 implementation on top of current main (conflict-free reverse-apply of the #123 hunks for the two `lsmt/file` files):

- Prune scan runs as a single `spawn_blocking` instead of one async round-trip per entry.
- Prune is triggered once per 1/8 of `max_dir_bytes` of new writes, at most one scan per directory at a time; growth charged during a scan chains a follow-up scan.
- The merged index is dropped before the lock-release/prune tail, so queued writes stop pinning it.

Later changes to the same file from #3522df8 / #e97064f (sys-module refactor, unaligned-extent rejection) are preserved — the restore only touches the premerged sections.

## Verification

- `cargo test -p overlaybd premerged` — 8 passed, including the restored prune-accounting tests from #179.
- `cargo fmt -p overlaybd -- --check` and `cargo clippy -p overlaybd --all-targets` clean.

### DIFF
--- a/storage/overlaybd/src/lsmt/file/helper.rs
+++ b/storage/overlaybd/src/lsmt/file/helper.rs
@@ -9,7 +9,7 @@ use std::io::{self, ErrorKind};
 use std::mem::size_of;
 use std::os::unix::fs::{FileExt, OpenOptionsExt};
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, OnceLock, Weak};
+use std::sync::{Arc, Mutex as StdMutex, OnceLock, Weak};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::sync::{Mutex, OwnedMutexGuard};
 use uuid::Uuid;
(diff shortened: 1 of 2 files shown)


yingdi-shan (approved):
Thanks for your contribution!

PR #246, opened 2026-09-04
template: report failed build steps by number, not instruction text

nblintao:
## What happens today

When a template build step fails, AgentENV fills `BuildStatusReason.step` with the instruction text (`"RUN apt-get update && …"`, `src/template/step_executor.rs`). The official e2b Python SDK resolves that field as an index into its locally recorded stack traces — `get_build_step_index` maps `"base"` → 0, `"finalize"` → last, and otherwise calls `int(step)` with no fallback (`packages/python-sdk/e2b/template/utils.py`). So any failed build crashes the SDK's error path:

```
ValueError: invalid literal for int() with base 10: 'RUN apt-get update && …'
```

…and the real build error (already carried in `message`) is never raised. The JS SDK doesn't crash but silently loses the traceback (`Number(step)` → `NaN`).

## The contract

The OpenAPI spec leaves `step` an unconstrained string ("Step that failed"), but e2b Cloud fills it with the **1-based step number** (`e2b-dev/infra`, `packages/orchestrator/pkg/template/build/phases/utils.go`: `strconv.Itoa(*StepNumber)`, numbered from 1 in `phases/steps/factory.go`), with `"base"`/`"finalize"` for those phases — and both official SDKs hard-code exactly that mapping. De-facto, numeric-or-sentinel is the contract.

## Change

- `reason.step` now carries the 1-based step number as a string.
- The instruction text moves into `message`, next to the exit status and stderr it already reports:
  ```
  message: "build step failed: RUN apt-get update …: command exited with status 100; stderr: …"
  step:    "3"
  ```
- Failures outside the step executor (base image, pipeline) keep `step: None`, which both SDKs skip safely.
- Existing test updated; new test pins the 1-based position of a failing RUN.

### DIFF
--- a/src/api/impls/template_helpers.rs
+++ b/src/api/impls/template_helpers.rs
@@ -72,8 +72,10 @@ pub(super) fn template_build_spec_from_start_request(
         spec = spec.ready_cmd(ready_cmd.clone());
     }
 
-    for step in body.steps.as_deref().unwrap_or_default() {
+    for (index, step) in body.steps.as_deref().unwrap_or_default().iter().enumerate() {
+        let appended_from = spec.step_count();
         spec = apply_e2b_template_step(spec, step)?;
+        spec = spec.stamp_source_steps(appended_from, index + 1);
     }
 
     Ok(spec)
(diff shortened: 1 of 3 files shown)


LSX-s-Software (approved):
Thanks for identifying and fixing the bug.

PR #252, opened 2026-09-07
fix(volume): avoid VM snapshots during sandbox deletion

yingdi-shan:
## What

Delete sandboxes with mounted volumes by freezing the writable filesystems, sealing and publishing their layers, and stopping the VM. Deletion no longer captures disposable memory, rootfs, or VM state.

## Why

The existing deletion path pauses the entire sandbox to preserve volume writes. A volume-only checkpoint avoids that work and leaves a clean filesystem journal for the next mount.

## Related issue

Extracted from #247 alongside #253 (OCI UID/GID startup) and #254 (BuildKit). The layer-reuse optimization in #249 remains separate.

## Scope and non-goals

Mounted-volume deletion, timeout eviction, failure recovery, regression tests, and volume documentation. BuildKit integration and OCI UID/GID resolution are submitted separately.

- Keep volume reservations until publication succeeds and the VM stops.
- Thaw and restore the original runtime after recoverable capture or publication failures. Terminal failures tear down the runtime and its persisted state.
- Bound freeze/thaw RPCs and check mountpoints before freezing, so an unmounted path cannot freeze the guest root filesystem.
- Explicit pause and graceful shutdown continue to save resumable VM state.

- Public API or generated protocol: unchanged; the internal sandbox backend gains freeze/thaw operations.
- Upgrade and rollback: no migration required.
- Host requirements, permissions, ports, or dependencies: existing Firecracker, ublk, and guest BusyBox requirements.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
make fmt clippy: passed
make test-unit: passed, including capability and ublk checks
cargo test -p agentenv --test integration frozen_volumes_survive_deletion_without_vm_snapshot -- --nocapture: passed
git diff --check: passed
```

The VM test passed with multiple volumes, recovery after a partial freeze failure, thaw/reseal, subPath mounts, a non-root default user, and dirty-file preservation in a replacement VM. It ran as a non-root user through the repository capability runner with isolated state and dependency copies. No timing benchmark was run; the regression tests verify that deletion does not invoke a full VM pause/snapshot. Go and generated API checks do not apply.

## Risks and reviewer notes

Review deletion ordering in `src/orchestrator/service.rs` and freeze recovery in `src/sandbox/firecracker/sandbox.rs`. Filesystem freezing flushes filesystem state; application buffers still require application-level coordination. Guest root-selection helpers overlap identically with the OCI startup fix and merge cleanly.

## Checklist

- [x] New behavior is covered by focused regression tests.
- [x] No generated code was manually edited.

### DIFF
--- a/src/orchestrator/tests.rs
+++ b/src/orchestrator/tests.rs
@@ -2,6 +2,7 @@ use std::collections::{HashMap, VecDeque};
 use std::net::Ipv4Addr;
 use std::path::{Path, PathBuf};
 use std::result::Result as StdResult;
+use std::sync::atomic::AtomicBool;
 use std::sync::{Arc, Mutex as StdMutex, OnceLock};
 
 use async_trait::async_trait;
@@ -108,6 +109,7 @@ fn make_orchestrator_without_background_with_factory_and_persister<
         factory,
         persister,
         sandboxes: RwLock::new(HashMap::new()),
+        deletions: tokio::sync::Mutex::new(HashMap::new()),
         proxy_routes: RwLock::new(ProxyRouteTable::default()),
         next_proxy_route_version: AtomicU64::new(1),
         counters: Default::default(),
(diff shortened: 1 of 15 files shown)


src/orchestrator/mod.rs
```diff
@@ -41,6 +41,7 @@ pub enum SandboxOperation {
     UpdateNetwork,
     PatchCustomExtensionParams,
     Stop,
+    Delete,
```
LSX-s-Software: `Stop` already means `Delete`. There's no need to distinguish them.

src/orchestrator/tests.rs
```diff
@@ -2715,6 +2716,270 @@ async fn orchestrator_list_empty_returns_empty() -> Result<()> {
...
+    let id = metadata.id;
+    orchestrator.delete_sandbox(id).await?;
+    assert_eq!(
+        *calls.lock().unwrap(),
+        vec![MockOperation::SnapshotVolumes, MockOperation::Stop]
+    );
+    assert!(orchestrator.get_sandbox(&id).await?.is_none());
+    assert!(orchestrator.sandboxes.read().await.get(&id).is_none());
+    Ok(())
+}
```
LSX-s-Software: This test is redundant and meaningless.

LSX-s-Software (changes requested):
The following are some of the problems that deserves solving.

p.s. The overall code quality of the volume feature is rather poor, but I don't think any of us have the time to review and optimize it. We'll just have to put them on hold for now.

LSX-s-Software (approved):

PR #256, opened 2026-09-07
feat(snapshot): compress layers at publish time when uploading to OSS/ACR

huajq:
## What

Redesign snapshot layer compression: local layers always stay raw, and compression happens once when layers are uploaded to OSS/ACR. A single new switch, `[snapshot.publish_compression]` (`enabled` / `algorithm` (`lz4`|`zstd`) / `workers`, on by default), replaces the legacy capture-time knobs — `[memory_snapshot].compression_*` and `[template_build]` are removed from the configuration schema.

## Why

Benchmarks show capture-time compression is a local pessimization: the guest working set is unchanged by compression (~62 MiB in the test setup), and while layer files are page-cache-warm the saved bytes never materialize — yet decompression CPU is paid on every fault (cold resume: raw 63.8 ms vs lz4 +18% vs zstd +106%; hot path identical at ~15 ms). The bytes compression actually saves only matter on the network path: OSS/ACR upload and cross-node resume.

## Related issue

Closes # (no tracking issue; motivation is the measurement analysis above)

## Scope and non-goals

- Included: publish-time compression in the OSS managed-layer and ACR source-registry upload paths (memory layers, rootfs read-write deltas, attached-drive deltas uniformly); removal of capture-time compression and the template-build compression override pipeline; P2P digest-key consistency guard; config, docs, and tests.
- Non-goals: no `Repository` trait changes; POSIX backend unchanged (local stays raw); zero read-path changes (ZFile is already transparent via switch-file magic detection); full P2P alignment for compressed digests (left as a TODO).

- Capture path: memory / rootfs / attached-drive layers are always sealed raw (daemon `close_seal` unchanged).
- Publish path: new shared helper `prepare_layer_upload` (`common/recontainerize.rs`) probes `is_zfile`, recontainerizes raw local layers with `compact_layers` into a temporary ZFile, then hashes and uploads the compressed bytes. `ManagedLayer.digest/size`, OSS object keys, and ACR blob annotations all reference the actually-uploaded bytes. Remote lowers are re-referenced untouched; already-ZFile inputs are passed through (idempotent).
- Resolution/resume: unchanged — consuming nodes auto-detect the format by content magic and need no configuration or flag.
- Failure handling: compression/hash failures abort the publish through the existing rollback path (`delete_prefix`); no partial uploads; temp files are cleaned up by guard.
- Known trade-off: ZFile layers record `uuid = None`, so P2P uuid-keyed acceleration does not apply to them (the daemon virtual-size probe falls back to a full `ImageFile` open, one-time).

- Configuration or defaults: new `[snapshot.publish_compression]` section (**enabled by default**, `lz4`) — OSS/ACR uploads start compressing on upgrade; set `enabled = false` to keep raw uploads. Legacy keys removed from the schema; unknown TOML keys are ignored by the config parser, so stale configs keep starting.
- Snapshot manifest, artifact layout, or storage format: when enabled, manifest digests/sizes reference compressed bytes; readers auto-detect by content magic, and existing repositories (raw or previously compressed layers) remain readable.
- Upgrade and rollback: rollback-safe — disabling the switch resumes raw uploads; no data migration needed. Note the default is **on**, so existing OSS deployments will upload compressed layers after upgrade (readers auto-detect; downgrade-safe).
- Host requirements, permissions, ports, or dependencies: none added.

## Validation

- [x] `make fmt` (`cargo fmt --all -- --check`)
- [x] `make clippy` (`cargo clippy -p agentenv -p agentenv-e2e-tests --all-targets -- -D warnings`)
- [x] `make test-unit` (`cargo test -p agentenv --lib`: 817/818; the single failure is a pre-existing flaky `api::proxy` test that passes in isolation, unrelated)
- [x] Relevant Rust integration tests (`fc::memory_snapshot_format_matches_config_and_resumes`, `fc::backend_pause_state_round_trips_through_encoded_artifacts`)
- [x] Documentation updated (AGENTS/CLAUDE.md, `reference.md`, `default.toml`, `p2p-design.md`)

Commands and results:

```text
cargo fmt --all -- --check                                    # pass
cargo clippy -p agentenv -p agentenv-e2e-tests --all-targets -- -D warnings  # pass
cargo test -p agentenv --lib                                  # 817 passed, 1 flaky (unrelated)

# OSS e2e (minio), incl. the new compression case
cargo test -p agentenv-e2e-tests --test snapshot_oss_e2e_test -- --include-ignored --test-threads=1
# 5 passed (incl. snapshot_oss_publish_compresses_raw_layers_when_enabled)

# KVM integration (real Firecracker pause/resume)
cargo test -p agentenv --test integration --release -- fc::memory_snapshot_format_matches_config_and_resumes fc::backend_pause_state_round_trips_through_encoded_artifacts
# 2 passed

# Real OSS end-to-end (bucket agentenv-oss-validation):
# publish (compression on): memory layer 53MB -> 34.7MB; fresh-home cross-node
# resume succeeded with guest data intact; cached layer verified as ZFile magic;
# guest runtime upper stays raw.
```

Performance comparison (real OSS, release build; template-build scenario, memory layer 87.8 → 31.9 MiB, 2.75×):

| Metric | raw | publish_compression (lz4) |
|---|---|---|
| Build publish phase (capture + compress + upload) | 2.28 s | **1.02 s (2.2× faster)** |
| Memory-layer upload time | 1.60 s | 0.49 s |
| Cold start from template (fresh node, OSS caches cleared, mean of 3) | 3.01 s (3.42/3.17/2.44) | **2.05 s (2.02/2.07/2.06), −32%** |
| Bytes downloaded during cold start | ~64 MiB | ~30 MiB |

Evidence for removing capture-time compression: cold resume raw 63.8 ms vs lz4 75.1 ms (+18%) vs zstd 131.4 ms (+106%); hot path identical at ~15 ms; guest working set (~62 MiB) is codec-independent.

Skipped checks and reasons: `make -C services test` (`services/` untouched); no generated-code changes.

## Risks and reviewer notes

- Core invariant: manifest digest/size must equal the actually-uploaded (compressed) bytes — guarded by unit tests in `recontainerize.rs` and the minio e2e. Please focus review on `common/recontainerize.rs` (new), the three upload branches in `oss/repository.rs`, and `acr/publisher.rs::upload_local_delta`.
- P2P: digest-keyed publication is skipped when the committed record names different bytes than the local descriptor (conservative; see the TODO in `p2p.rs` about propagating compressed paths to restore acceleration).
- ZFile layers get `uuid = None`: P2P uuid-keyed acceleration is unavailable for them (pre-existing graceful degradation, not a regression).
- Benchmarks ran on a shared test host, so absolute numbers carry some noise; the compressed runs show noticeably tighter variance (2.02–2.07 s vs 2.44–3.42 s) because less of the critical path is network-exposed.

## Checklist

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -231,6 +231,18 @@ p2p_enabled = true
 # [snapshot.image_publish]
 # enabled = true
 
+# Compress memory layers and incremental read-write layers once, when they are
+# uploaded to OSS/ACR. Local layers always stay raw, so local resume pays no
+# decompression cost; enabled by default to cut network bytes for cross-node
+# resume. Disable when upload CPU matters more than transfer bytes.
+[snapshot.publish_compression]
+enabled = true
+# Supported algorithms are "lz4" and "zstd".
+algorithm = "lz4"
+# Workers compressing 4 KiB blocks. One is sequential; higher values run in
+# parallel without changing the output layout.
+workers = 1
+
 # POSIX storage defaults under AENV_HOME. Set this for a shared or independent
 # snapshot filesystem.
 # [backend.posix_fs]
@@ -239,8 +251,8 @@ p2p_enabled = true
 # OSS requires an endpoint, bucket, region, and either credential_process or
 # static credentials. credential_process takes precedence over static values.
 # [backend.oss]
-# endpoint = "https://oss-cn-beijing-internal.aliyuncs.com"
-# bucket = "agentenv-snapshots"
+# endpoint = "https://<your-oss-endpoint>"
+# bucket = "your-snapshot-bucket"
 # region = "cn-beijing"
 # Bucket addressing style: "virtual" or "path". Leave unset to use
 # endpoint-based auto-detection.
@@ -317,15 +329,6 @@ fill_concurrency = 4
 # OverlayBD path. Disable with AGENTENV_MEMORY_SNAPSHOT_TRACK_DIRTY_PAGES=false.
 track_dirty_pages = true
 
-# Compress memory snapshot layers. The algorithm is parsed but ignored when
-# compression is disabled.
-compression_enabled = false
-# Supported algorithms are "lz4" and "zstd".
-compression_algorithm = "lz4"
-# Workers compressing 4 KiB blocks. One is sequential; higher values run in
-# parallel without changing the output layout.
-compression_workers = 1
-
 [memory_snapshot.background_download]
 # Download missing blocks from remote memory layers; never start before envd
 # becomes ready.
@@ -346,19 +349,6 @@ concurrency = 4
 # block_size must not exceed 2 GiB; defaults use 256 MiB. Per-image overrides
 # cannot resize it.
 max_inflight_blocks = 16
-
-[template_build]
-# Compress snapshot artifacts captured by template builds, independent of
-# [memory_snapshot].compression_enabled (template builds consult only this
-# section). When enabled, both memory layers and the sealed rootfs read-write
-# layer are compressed. Pause/snapshot captures of running sandboxes are not
-# affected.
-compression_enabled = false
-# Supported algorithms are "lz4" and "zstd".
-compression_algorithm = "lz4"
-# Workers compressing 4 KiB blocks. One is sequential; higher values run in
-# parallel without changing the output layout.
-compression_workers = 1
 
 [observability]
 # Expose node identity, host metrics, and orchestrator runtime metrics.
--- a/crates/e2e-tests/tests/snapshot_oss_e2e_test.rs
+++ b/crates/e2e-tests/tests/snapshot_oss_e2e_test.rs
@@ -1,7 +1,10 @@
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, OnceLock};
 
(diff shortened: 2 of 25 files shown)


config/default.toml
```diff
@@ -231,6 +231,18 @@ p2p_enabled = true
...
+# uploaded to OSS/ACR. Local layers always stay raw, so local resume pays no
+# decompression cost; enabled by default to cut network bytes for cross-node
+# resume. Disable when upload CPU matters more than transfer bytes.
+# [snapshot.publish_compression]
+# enabled = true
+# Supported algorithms are "lz4" and "zstd".
+# algorithm = "lz4"
+# Workers compressing 4 KiB blocks. One is sequential; higher values run in
+# parallel without changing the output layout.
+# workers = 1
```
yingdi-shan: Please enable compression by default.

yingdi-shan (changes requested):

config/default.toml
huajq: already done.

yingdi-shan: OK. I think we should uncomment the configuration. Otherwise the configuration may seem confused to users.

yingdi-shan (approved):

PR #265, opened 2026-09-08
fix(sandbox): clean up empty serial logs and expire inactive logs

### LINKED ISSUE #260: Sandbox create throughput halves as the serial log directory grows
### Affected component or operation

Sandbox creation / boot

### Regression summary

The persistent serial-output root, `{serial_dir}/` (default `$AENV_HOME/logs/serial`), receives one subdirectory per sandbox and nothing ever prunes it. Sandbox create throughput on a node therefore decays with the node's uptime: every create links a new entry into a directory that only grows.

Measured on one node under sustained create load: at ~594k entries in the root it sustained 197–280 creates/sec. With the root emptied and nothing else changed, the same node sustained 460–475 creates/sec at 76–78% CPU — **~1.9×**. Under that load the root grew at ~470 entries/sec.

Nearly every one of those subdirectories holds a single zero-byte `firecracker-stderr.log`: a guest resumed from a snapshot writes nothing to the console. So the cost is not in writing serial output, it is in linking the directory entry under an exclusive lock on the parent inode — and it falls on every create, whether or not anyone ever reads a log.

User impact: a node silently loses roughly half its create throughput as it ages, with no error, no log line, and no metric pointing at the cause. It presents as general degradation.

### Compared versions

- Baseline: `2d766447261a7252447fe8bccef2e87d48ece5a6`
- Current: `53e4061bed1e8a6d0205550e20274a68311e1fe5`
- First known bad: n/a
- Local modifications: -

### Host and workload

- Linux x86_64, bare metal, KVM.
- Sandboxes created from snapshots (hence the empty console output), ublk + OverlayBD storage path.
- Sustained concurrent create load against a single node; same workload generator, concurrency, and duration on both sides of the comparison.
- `{serial_dir}` set to a persistent host path, which is the configuration the docs describe; local ext4-family filesystem.

### Benchmark command and procedure

```shell
1. Drive sustained concurrent sandbox creates at a single node; record creates/sec and host CPU once the rate is steady.
2. Count the serial root (`ls -U {serial_dir} | wc -l`) and note its directory inode size (`stat {serial_dir}`).
3. Move the root aside rather than deleting in place: `mv {serial_dir} {serial_dir}.parked && mkdir {serial_dir}`, then delete the parked copy out of band, after measuring. 
4. Repeat step 1 with the same generator, concurrency, and duration.

Both measurements are steady-state windows on the same node, minutes apart,
with no restart or configuration change between them.
```

### Results

| Serial root   | Create throughput | Host CPU |
|---------------|-------------------|----------|
| ~594k entries | 197–280 /sec      | —        |
| empty         | 460–475 /sec      | 76–78%   |

Throughput was recorded as a steady-state range over the measurement window rather than as a percentile distribution, so treat these as ranges, not distributions.

The inverse correlation between root size and peak create rate held across all four nodes in the fleet at the time the problem was found (size column is disk usage of the tree):

| Node | Entries in root | Tree size | Peak creates/sec |
|------|-----------------|-----------|------------------|
| A    | 593,885         | 4.6 G     | 408              |
| B    | 579,462         | 4.5 G     | 428              |
| C    | 351,970         | 2.8 G     | 457              |
| D    | 347,344         | 2.7 G     | 452              |

### Raw data

```text
Counts on the same four nodes a while later, showing that growth does not level off (this column is the directory inode size, not tree size):


node B   650,025 entries   41.6 MB directory inode
node C   411,348           26.5 MB
node D   408,246           26.4 MB
node A    58,596            3.8 MB   (root had been parked earlier)
```

### Profiles and diagnostics

_No response_

### Measurement checklist

- [x] I ran the baseline and current versions on the same host or documented the differences.
- [x] I repeated the measurement and reported the sample count.
- [x] I controlled or documented cold/warm caches and other relevant variables.
- [x] I searched existing issues and did not find a duplicate.

yingdi-shan:
## What

Remove empty managed Firecracker logs after confirmed process exit and expire inactive nonempty logs after a configurable retention period. Keep the existing `{serial_dir}/{sandbox_id}/` layout.

## Why

Silent snapshot launches leave permanent directories and empty log files behind, growing the serial root with every sandbox created. Cleaning stopped sandboxes prevents this accumulation during normal operation.

## Related issue

Closes #260. Alternative to #261.

## Scope and non-goals

Includes cleanup after stop, retention, live-writer protection, documentation, and regression tests. No sharding, lazy capture, active-log rotation, or throughput improvement claim. Legacy directories and logs left after an abnormal drop or server crash require offline cleanup.

- Hold a directory lock before opening or relocating managed logs. Lock contention yields asynchronously and can be cancelled.
- After stop confirms process exit, create an `.inactive` marker under the lock and remove empty known logs. Resuming an ID clears the marker before opening logs. Abnormal drop leaves logs unmarked, preserving late internal-logger output even with custom stdout/stderr paths.
- The background collector only handles marked sandbox-ID directories, taking a nonblocking exclusive directory lock. It preserves symlink entries and custom files, and never recursively deletes directories or scans host process descriptors.
- Start a sweep with the server and repeat 60 seconds after each pass. Filesystem work runs on the blocking pool; shutdown cancels between directory entries and joins the worker. Entries disappearing during a sweep are skipped.
- Retain inactive nonempty logs for `firecracker.serial_log_retention_secs`, default 604800 (seven days). Zero disables expiry. Retention uses the newest directory/log modification time, refreshed after confirmed exit.
- Preserve startup diagnostics and append to retained stdout/stderr on resume by bypassing the warm process pool when necessary.
- Keep process-wide networking available across individual orchestrator shutdowns. The server shuts it down after sandboxes and the Firecracker pool; the restart integration test exercises a fresh process and retained stderr history.

- Public API or generated protocol: no HTTP or schema changes. Adds a Rust process-shutdown helper; an individual orchestrator shutdown no longer stops shared networking permanently.
- Configuration or defaults: adds `firecracker.serial_log_retention_secs` and `AENV_FIRECRACKER_SERIAL_LOG_RETENTION_SECS`, default seven days.
- Snapshot manifest, artifact layout, or storage format: snapshot formats unchanged. Managed serial directories may contain an internal `.inactive` marker after stop.
- Upgrade and rollback: unmarked legacy/crash logs are preserved. Clean those manually only after stopping all servers and Firecracker processes using the root. Do not share a serial root with older servers that do not honor the marker protocol. Rollback stops collection but cannot restore expired logs.
- Host requirements, permissions, ports, or dependencies: no new dependencies. The serial root must be node-local and writable only by the trusted runtime user. Explicit stdout/stderr overrides outside the managed tree are not collected.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
cargo test -p agentenv --lib serial --no-fail-fast
  24 passed
make fmt
  passed
make clippy
make test-unit AENV_TEST_STATE_DIR=/tmp/aenv-serial-retention-tests AENV_TEST_DEPS_PATH=/tmp/aenv-serial-retention-tests/deps
  passed: 839 agent tests, 4 capability tests, 3 linux-cap tests,
  8 ublk tests, 46 daemon tests, and both shell verification scripts
make test-agent-integration PROFILE=debug AENV_TEST_STATE_DIR=/tmp/aenv-serial-retention-tests AENV_TEST_DEPS_PATH=/tmp/aenv-serial-retention-tests/deps
  passed: 18 VM integration tests, 2 orchestrator tests, 5 snapshot OSS tests
```

All builds and Rust tests run as the non-root workspace user. Test state and dependencies are isolated under `/tmp/aenv-serial-retention-tests`; privileged tests use the repository capability runner.

Skipped checks and reasons:

- Go tests and API regeneration: no Go, schema, or generated source changes.
- Broader envd/storage integration suites: not rerun; this change affects Firecracker log lifecycle.
- Throughput benchmarks: no performance measurements are claimed.

## Risks and reviewer notes

The main review area is `src/sandbox/firecracker/serial_logs.rs`: eligibility after confirmed exit, directory locking, and retention. Regression coverage includes late unlocked logger writes after abnormal drop, interrupted resume, cancellable lock contention, collection during repeated launches/stops, symlinks/custom entries, startup failure, retained history, and warm-log relocation.

Unmarked logs need offline cleanup. Retention bounds eligible inactive-log lifetime; it does not limit active-log size, peak directory count, or compact an already expanded filesystem directory inode. Filesystem syscalls themselves cannot be interrupted by worker cancellation.

## Checklist


makhov:
The failure mode in #260 isn't cumulative growth per se — it's the cost of inserting into and looking up in a huge parent directory, paid by every create. There are three paths to the same degraded state intact:                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                                                                                           
- the growth is unbounded for live sandboxes
- the directory inode never shrinks. So one overload burst permanently bloats the root inode even after cleanup empties it
- crashes and abnormal drops leak forever and then see the previous point                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                                           
Other notes worth mentioning:
                                                                                                                                                                                                                                                                                                                                                                                                                                           
- it claims "Closes #260" with no benchmarks run                                                                                                                            
- the sweep is a recurring full read_dir over the root, with ~6 syscalls per entry (open + flock + marker stat + up to 3 log stats), every 60 seconds. On a large root that's a periodic I/O storm competing for the same parent inode locks as creates — it can add contention in exactly the regime #260 measures 
- resume loses warm-pool reuse when retained logs exist (sandbox.rs: the new !firecracker_stdout_path().exists() guard forces a fresh Firecracker spawn)                                                                                                                                                                                                                                                                                                                                                                                            
- scope creep: the PR also changes orchestrator/network shutdown semantics                                                                                                                     

I think the cleanup can be a nice addition to #261, which fixes the issue, not masking the symptoms

yingdi-shan:
Thanks for the detailed feedback. Our production workload generally uses long-lived sandboxes, so the creation rate in #260 is not representative of our primary use case.

Our preferred direction is to delete obsolete logs and reclaim their directories so historical sandbox count does not permanently affect the node. Sharding would distribute the entries, but the log tree would still grow without bound, so we do not see it as the ultimate solution.

Another solution is to combine logs from different sandboxes into a single log, which makes it easier to inspect serial output while also improving performance and simplifying log size control.

PR #266, opened 2026-09-09
fix: wait for daemon exit during server shutdown

### LINKED ISSUE #158: First sandbox start after 'systemctl restart aenv' fails with ublk-daemon.sock: No such file or directory
### Summary

After `systemctl restart aenv`, the first sandbox start reliably fails with:

```
ublk-daemon.sock: No such file or directory
```

Restarting the service a second time clears it, and subsequent starts work normally. It looks like a socket-unlink race between the outgoing daemon and the incoming one rather than a configuration problem, but the failure surfaces as a confusing "file not found" at sandbox-create time rather than at service start.

Filing this because we have hit it consistently enough to write "if this happens, restart once more" into our own runbook, and that felt like something upstream would rather know about than have operators quietly work around.

### Environment

- AgentENV v0.1.1, built from source
- Ubuntu 26.04, systemd-managed `aenv.service`, non-root service account with KVM access via supplementary group
- Reproduces on a single-node install

### Reproduction

1. `systemctl restart aenv`
2. Immediately: `aenv start -d --timeout 600 ubuntu`
3. First attempt fails with the error above
4. `systemctl restart aenv` again
5. Same start command now succeeds

The timing window matters — if enough time passes between the restart and the first start, we do not see it. That is what makes us think it is a race on the socket path rather than a persistent state problem.

### What we have not done

We have not instrumented the daemon to confirm the ordering, so "socket-unlink race" is inference from the symptom and the timing sensitivity, not something we have proven. Happy to gather more if it would help — journal excerpts around the restart, strace on the daemon, or a tighter reproduction script.

### Why it is worth fixing even though the workaround is trivial

The workaround is a second restart, so the operational cost is low. The reason to fix it is that the error names a socket path, which sends operators looking for a missing file or a permissions problem in their own install rather than at a startup ordering issue. Either waiting for the socket to be ready before accepting sandbox creates, or returning a "daemon still starting" error rather than a filesystem error, would remove that dead end.

Related but distinct from the existing ublk issues (#93 kernel warnings, #130 kernel 6.6 zero-copy) — this one is about restart ordering, not kernel support.

zlzGithub-0801:
## What

Wait for the ublk daemon process to fully exit before completing server shutdown.

## Why

The daemon may acknowledge shutdown before finishing cleanup. If the server restarts during this window, the old daemon can remove the new daemon's Unix socket, causing subsequent requests to fail.

## Related issue

Closes #158

## Scope and non-goals

This change only coordinates shutdown between the server and its managed ublk daemon. It does not add daemon auto-restart or change systemd configuration.

The daemon watchdog publishes a retained exit notification after `child.wait()` completes. `shutdown()` sends the shutdown request and then waits for that notification.

- Upgrade and rollback: N/A.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [ ] Documentation updated

Skipped checks and reasons:

No generated code, services, documentation, or performance-sensitive behavior changed.

## Risks and reviewer notes

The server shutdown now waits until the daemon process has actually exited.

## Checklist

### DIFF
--- a/storage/ublk-daemon/src/client.rs
+++ b/storage/ublk-daemon/src/client.rs
@@ -129,6 +129,8 @@ struct UblkDaemonClientInner {
     socket_path: PathBuf,
     /// Set to `true` when the daemon process exits (expected or unexpected).
     daemon_dead: AtomicBool,
+    /// Retains completion so shutdown callers cannot miss the watchdog's exit.
+    daemon_exited: tokio::sync::watch::Sender<bool>,
     /// Set to `true` when `shutdown()` is called, so the watchdog task
     /// doesn't log the expected exit as an error.
     shutting_down: AtomicBool,
@@ -224,6 +226,7 @@ impl UblkDaemonClient {
             inner: Arc::new(UblkDaemonClientInner {
                 socket_path: config.socket_path,
                 daemon_dead: AtomicBool::new(false),
+                daemon_exited: tokio::sync::watch::channel(false).0,
                 shutting_down: AtomicBool::new(false),
                 death_reason: Mutex::new(None),
                 runtime_device_timeout: config.runtime_device_timeout,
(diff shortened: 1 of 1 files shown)


yingdi-shan (approved):
LGTM.

PR #267, opened 2026-09-09
feat(snapshot): prefetch envd memory pages at resume for faster envd-ready

huajq:
## What

Record the guest init daemon's (envd) resident guest-physical address ranges at snapshot capture and bulk-prefetch those pages into the node's remote-block cache before the VM resumes, so envd's early startup no longer waits on one-at-a-time remote reads.

## Why

Resuming a snapshot stalls envd-ready on a serial demand-page-fault chain (~256 KiB × ~21 ms OSS RTT per read, bounded by vCPU count). Measured on a cold OSS-backed snapshot: envd-ready improves from **5.47 s to 2.13 s (-61%)** with prefetch enabled.

## Related issue

Closes #

## Scope and non-goals

- Included: GPA-range export via a new `aenv-pagedump` tools-drive binary; `memory-prefetch.json` artifact publish/resolve on OSS and posix_fs; a `Prefetch` RPC on the ublk daemon; resume-time trigger; tests and docs.
- Excluded (non-goals): KVM pre-fault / `KVM_PRE_FAULT_MEMORY` (requires kernel ≥ 6.11, not available on the fleet); page-fault trace recording/replay; changes to snapshot layering or capture content; any modification of the guest kernel.

```
capture: tools-drive aenv-pagedump reads /proc/<envd-pid>/maps+pagemap
         → GPA ranges → artifacts/{id}/memory-prefetch.json (best-effort)
resume:  resolver downloads manifest (404 tolerated) → start_resume triggers
         ublk-daemon Prefetch RPC (fire-and-forget) → 6 workers × 4 MiB chunks
         via ImageFile::read_at → remote-block cache warmed before guest runs
```

- Strictly best-effort at every hop: missing binary, missing/invalid/oversized manifest, or RPC failure all fall back silently to the existing on-demand path; resume never fails because of prefetch.
- Manifest stores GPA ranges only, so chained snapshots (delta layers) stay valid.
- Export runs inside a bounded `spawn_blocking` + dedicated runtime with a 10 s timeout so a wedged envd can never stall capture.
- Older tools drives (no `aenv-pagedump`) simply produce no manifest — behavior unchanged from today.

- Public API or generated protocol: additive `Prefetch` variant on the internal ublk-daemon Unix-socket protocol; old client → new daemon fails safe (warn only); new client → old daemon unaffected. No external API change.
- Configuration or defaults: N/A — feature is always on, no new config keys.
- Snapshot manifest, artifact layout, or storage format: adds one optional fixed artifact `artifacts/{id}/memory-prefetch.json`; `FirecrackerSnapshotManifest` gains a `#[serde(skip)]` runtime-only field; both are backward compatible.
- Upgrade and rollback: safe — snapshots with or without the manifest resume identically; rolling back simply stops producing/consuming manifests.
- Host requirements, permissions, ports, or dependencies: tools drive must be rebuilt to include `aenv-pagedump` (new `tools-image/pagedump` crate + Dockerfile stage); older tools drives degrade gracefully.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [x] `make test-unit`
- [ ] `make -C services test` (required when `services/` changes) — N/A
- [x] Documentation updated

Commands and results:

```text
cargo test -p agentenv --lib         → 838 passed, 0 failed
cargo test -p uvm-ublk-daemon        → 48 + 37 passed, 0 failed
cargo fmt --all -- --check           → clean
cargo clippy --all-targets -D warnings → clean
E2E on OSS validation bucket (cold resume):
  without prefetch: envd-ready 5.47 s | with prefetch: 2.13 s (-61%)
  export path verified in guest: "exported via aenv-pagedump binary"
  capture export overhead: ~0.68 s (shell pipeline) → ~0.1 s (tools-drive binary)
```

Skipped checks and reasons: `make -C services test` — no `services/` changes. Generated-code regeneration — N/A (hand-written protocol variant, no codegen involved).

## Risks and reviewer notes

- Capture makes bounded (10 s timeout) guest RPCs before pause — a new dependency on envd liveness at capture time; mitigated by the timeout and silent fallback.
- Prefetch is bounded by MAX_PREFETCH_RANGES (4096) / MAX_PREFETCH_BYTES (256 MiB); over-cap manifests are disabled, never truncated.
- GPA ranges equal memory-image offsets only for single-region guests (mem ≤ 3328 MiB); documented as best-effort for larger guests.
- Key files: `src/sandbox/firecracker/pagedump.rs`, `src/snapshot/prefetch.rs`, `storage/ublk-daemon/src/server.rs` (`handle_prefetch`), `tools-image/pagedump/src/main.rs`.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ env/
 
 ### Node.js
 node_modules/
+.worktrees/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,13 @@ implementation rather than after mistakes.
   to memory, rootfs, attached-drive, and volume layers. Capture leaves the source
   running; launching it creates a new sandbox. Pause/resume retains the existing
   ID through separate paused-sandbox persistence. Template builds commit snapshots.
+  At capture, the guest init daemon's (envd) resident guest-physical memory
+  ranges are also exported and published as a `memory-prefetch.json` artifact
+  next to `vm_state.bin`; at resume, AgentENV bulk-prefetches those pages into
+  the node's remote-block cache via the ublk daemon so the init path does not
+  stall on one-at-a-time remote reads. The prefetch is strictly best-effort:
+  missing, unreadable, or oversized manifests are ignored and resume always
+  falls back to the on-demand path.
 - **Volume:** An independently managed block filesystem that survives sandbox
   deletion. `exclusive` permits one writable mount; `ro` permits shared read-only
   mounts. `volumeMounts` references managed volumes; cold-start `attachedDrives`
--- a/docs/src/concepts/snapshots.md
+++ b/docs/src/concepts/snapshots.md
@@ -44,6 +44,25 @@ curl -H 'X-API-Key: test-key' \
   http://127.0.0.1:8000/snapshots/<snapshot-id-or-alias>
 ```
 
+## Resume-Time Memory Prefetch
+
+When a snapshot is created, AgentENV also records which guest-physical memory
+pages the sandbox's init process (envd) had resident, and stores the list as a
+small `memory-prefetch.json` artifact next to `vm_state.bin` in the snapshot
+repository. When a sandbox is later started from that snapshot, AgentENV
+bulk-prefetches those pages into the node's local remote-block cache before the
+VM resumes, so the init process and early startup path no longer wait on
+one-at-a-time remote reads.
+
+The prefetch is strictly best-effort:
+
+- Snapshots created before this feature (or from guests where the init process
+  cannot be inspected) simply resume through the normal on-demand path.
+- A missing, unreadable, or oversized manifest is ignored; resume never fails
+  because of it.
+
+No configuration is required.
+
 ## Use a Snapshot Rootfs as an OCI Image
 
 Starting with `aenv start <snapshot>` restores the complete snapshot state. If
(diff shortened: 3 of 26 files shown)


huajq:
Thanks for the review! Addressed in 8ac10b25:

**Fixed**
- Export now runs the tools-drive binary directly by process name (no `sh`/`pgrep` dependency) with `ProcessOpts` timeout, so a stalled dump is killed instead of lingering into the snapshot.
- Export-side sum uses `checked_add`; runtime-build failure falls back to `None` instead of `expect` panic.
- GPA ranges are translated to memory-image file offsets for multi-region guests (region 2 at GPA 64 GiB laid out sequentially after region 1), with unit tests.
- Manifest parsing normalizes (sorts and coalesces) ranges; artifact read is size-bounded; daemon-side RPC validation (range count / total bytes / checked arithmetic) added; the per-image read guard is now held for the whole background prefetch.
- Publish-retry no longer leaves a stale `memory-prefetch.json` when the new attempt has no source (both OSS and posix_fs backends).
- posix_fs resolver existence check is async (`try_exists`).
- tools-image: `ARG GOPROXY` declared, publish recipe gets the same proxy args, and proxy values are no longer echoed in make recipes (name-only build args).

**Reviewed but intentionally not changed**
- *Completion barrier for prefetch*: deliberate design. Prefetch is fire-and-forget and purely additive — demand reads always work; prefetch just wins the race for most pages. A barrier would add a hard floor equal to the whole prefetch duration even if envd only needs the first few MiB.
- *`.dockerignore` `!pagedump`*: verified by a successful build — `aenv-pagedump` is present in the produced `tools.ext4`, so the current pattern does include the directory contents.
- *Daemon-wide prefetch concurrency bound*: noted as future hardening; current node-level OSS capacity plus the existing gate's FG accounting make this a theoretical concern at our resume rates.
- *Cross-arch (arm64) pagedump link*: our fleet is amd64-only; the stage stays on BUILDPLATFORM for now.

PR #272, opened 2026-09-10
fix(sandbox): honor disabled Firecracker logging

### LINKED ISSUE #260: Sandbox create throughput halves as the serial log directory grows

yingdi-shan:
## What

Honor disabled Firecracker logging for stdout/stderr capture on fresh boots, snapshot restores, and warm process launches. With `firecracker.log_level` unset or empty, launches send stdout/stderr to `/dev/null` and create no log files or per-sandbox serial directories.

## Why

The existing setting disables `firecracker.log`, but stdout/stderr files are still opened unconditionally. Silent snapshot launches therefore leave empty persistent directories behind. Issue #260 reports declining sandbox-create throughput as this directory grows.

## Related issue

Closes #260.

Related proposals: #261 (directory sharding) and #265 (log retention).

## Scope and non-goals

Covers logging defaults, cold and warm spawn paths, regression tests, and documentation. Does not add sharding, retention, or cleanup of existing logs.

- A non-empty log level enables stdout/stderr capture alongside `firecracker.log`. Setting `serial_dir` alone does not enable capture. Whitespace-only levels also disable logging.
- Explicit Rust stdout/stderr destinations enable only the requested streams even without a log level.
- The process pool opens logs only when capture is enabled and declines requests with different capture settings, since process stdio is fixed at spawn. Disabled warm capture skips log relocation.
- The integration test that inspects guest kernel messages now explicitly requests stdout capture.

- Configuration or defaults: no new keys. `firecracker.log_level` now controls default serial capture as well as the internal logger. Set it to `Info` to retain boot diagnostics and Firecracker logs.
- Upgrade and rollback: deployments without a log level stop collecting new serial logs. Existing files are preserved. Rollback restores unconditional serial capture.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [ ] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
make fmt
  passed
CARGO_TARGET_DIR=/home/ubuntu/AgentENV/target \
AENV_HOME_PATH=/tmp/aenv-issue-260-pr-tests/home \
AENV_RUNTIME_PATH=/tmp/aenv-issue-260-pr-tests/run \
AENV_DEPS_PATH=/tmp/aenv-issue-260-pr-tests/deps \
cargo test -p agentenv --lib sandbox::firecracker::
  78 passed
CARGO_TARGET_DIR=/home/ubuntu/AgentENV/target make clippy
git diff --check
  passed
```

The disabled-logging regression failed before the fix because spawning created a log directory. Coverage includes unset/empty/whitespace levels, persistent and work-directory destinations, cold and warm capture, enabled output and relocation, and explicit stream overrides. Builds and tests run as the non-root workspace user with isolated test state.

Skipped checks and reasons:

- Full `make test-unit` and VM integration suites: validation is focused on Firecracker process output and filesystem behavior; VM boot and the modified volume integration test were not rerun.
- Go tests and code generation: no Go or schema changes.
- Throughput benchmark: not rerun; no measured performance improvement is claimed.

## Risks and reviewer notes

Default startup errors no longer include captured stderr; enable logging when boot diagnostics are needed. Enabled persistent logs still require operator-managed retention, and existing directories are not pruned. Review the capture selection in `sandbox.rs` and pool compatibility check in `pool.rs`.

## Checklist


guozy18 (approved):
LGTM.

Issue #274, opened 2026-09-10
OverlayBD direct I/O hard-codes 512-byte alignment and fails with EINVAL on 4096-byte-sector block devices


ethe:
### Affected component

OverlayBD / image resolution

### AgentENV version

- Version or commit: `main` at b0d649c41112217ae3354eb189a6fb9996e5529b (2026-09-10)

### Host environment

- OS: Ubuntu 24.04 (Depot CI runner image `depot-ubuntu-24.04`)
- Kernel (`uname -a`): `6.12.8+` x86_64
- Architecture: x86_64
- Bare metal, VM, or cloud: cloud VM (Depot-hosted GitHub Actions runner), KVM available
- ublk support: yes
- CPU: 8 vCPU
- Memory: 31 GiB
- Storage/filesystem: root is ext4 on `/dev/mapper/root` (dm thin pool over virtio disks `vda`/`vdb`). The block device reports **4096-byte logical and physical sectors**:

  ```
  $ lsblk -t
  NAME       ALIGNMENT MIN-IO  OPT-IO PHY-SEC LOG-SEC ROTA SCHED       RQ-SIZE   RA WSAME
  vda                0 262144 2097152    4096    4096    1 mq-deadline     256 4096    0B
  └─thinpool         0 262144 2097152    4096    4096    1                     4096    0B
    └─root           0 262144 2097152    4096    4096    1                     4096    0B
  $ findmnt -no SOURCE,FSTYPE,OPTIONS --target /tmp
  /dev/mapper/root ext4   rw,noatime,lazytime,nobarrier
  ```

### Operation and configuration

```shell
# Test suite (TMPDIR defaults to /tmp on the 4096-byte-sector filesystem)
make test-ublk PROFILE=debug

# Minimal syscall-level reproduction of the same failure, no AgentENV involved
fio --name=p --directory=/tmp --ioengine=io_uring --direct=1 --bs=512 --size=1M --rw=write --iodepth=4
fio --name=p --directory=/tmp --ioengine=libaio   --direct=1 --bs=512 --size=1M --rw=write --iodepth=4
fio --name=p --directory=/tmp --ioengine=psync    --direct=1 --bs=512 --size=1M --rw=write --iodepth=4
# The same commands with --bs=4096 succeed.
```

[...]

PR #279, opened 2026-09-11
fix(overlaybd): derive direct I/O alignment from each file

ranxi2001:
## What

Query each direct-I/O file’s memory and offset alignment with Linux STATX_DIOALIGN and use those limits for LocalFile reads and writes, including io_uring. This allows sub-sector reads from local lower layers on filesystems requiring 4096-byte direct I/O.

## Why

The fixed 512-byte alignment causes EINVAL when the backing file requires 4096-byte requests. The libaio image path opens local lower layers with direct I/O, so even reading their headers can fail.

## Related issue

Related to #274. This addresses lower-layer reads and aligned LocalFile I/O; writable LSMT operations that issue 512-byte requests on a 4K-required file remain unsupported.

## Scope and non-goals

The change is limited to LocalFile and its existing platform syscall boundary. The LSMT data/index layout remains unchanged; it does not add read-modify-write or buffered fallback.

Alignment is queried once using the opened descriptor. Read ranges use the reported offset alignment, while bounce buffers use the memory alignment. Writes still require aligned offsets and lengths, and invalid requests report the required alignment. If the kernel/filesystem cannot report the limits, or on macOS, the existing 512-byte behavior is retained. Other query errors are propagated.

- Configuration or defaults: unchanged; buffered files do not query alignment.
- Upgrade and rollback: no migration; rollback restores the old alignment assumption.
- Host requirements, permissions, ports, or dependencies: no new dependency or privilege; STATX_DIOALIGN is used when available (Linux 6.1+ with filesystem support).

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [ ] `make test-unit`
- [ ] Documentation updated

Commands and results:

```text
cargo test --locked -p overlaybd --lib --all-features
  359 passed, 4 ignored
cargo clippy --locked -p overlaybd --all-targets --all-features -- -D warnings
  passed
make fmt
  passed
make clippy
  passed with locally extracted libclang and explicit resource directory
git diff --check
  passed
```

Full workspace clippy passed after supplying libclang/LLVM and Clang builtin headers from locally extracted distribution packages; no host packages were installed.

The no-default-features library suite also passed (299 tests, 2 ignored). A process-local syscall probe reporting and enforcing 4K constraints makes the same new lower-read regression fail on base with EINVAL and pass with this patch. The libaio image-open test passes under that probe. This models alignment constraints; it is not native 4K hardware validation, and it does not intercept io_uring submissions.

Skipped checks and reasons:

- Native 4K validation and macOS execution were unavailable.
- The new explicit context/io_uring test was attempted but failed before I/O during ring setup on local WSL 5.15 (EINVAL); the existing ring builder requires newer kernel setup flags.
- Privileged ublk, VM and full E2E tiers were not run on this host. Go, generated-code, documentation and benchmark checks are unrelated to this change.

## Risks and reviewer notes

This is a partial fix for #274. LSMT uses 512-byte data and seal-index/padding writes, which cannot be submitted as direct I/O to a file requiring 4096-byte requests. Such writes now fail with the required alignment in the error. Supporting them needs a separate storage design. Filesystems that cannot report STATX_DIOALIGN retain the existing 512-byte assumption.

## Checklist

### DIFF
--- a/storage/overlaybd/src/backend/local.rs
+++ b/storage/overlaybd/src/backend/local.rs
@@ -433,7 +433,14 @@ impl LocalFileInner {
         // For O_DIRECT, we must use aligned memory.  io_uring still requires
         // aligned buffers for O_DIRECT files, so we submit the aligned read
         // and then slice out the requested region.
-        let got = io_ring::read_exact_at(submitter, fd, buffer.as_mut(), aligned_offset).await?;
+        let got = io_ring::read_exact_at_aligned(
+            submitter,
+            fd,
+            buffer.as_mut(),
+            aligned_offset,
+            self.alignment.memory,
+        )
+        .await?;
 
         if got <= head {
             return Ok(Bytes::new());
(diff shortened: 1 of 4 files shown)


ranxi2001:
Thanks for catching these issues. Fixed in `f43c506` and `e357a2d`: `statx` alignment values are now validated for zero, overflow, and power-of-two constraints. Direct-I/O io_uring reads and writes now use alignment-aware completion handling and stop before retrying with an unaligned buffer after a short completion. OverlayBD format, clippy, and unit tests pass (`359 passed, 4 ignored`). The remaining ignored io_uring context test requires a newer host kernel than this WSL environment.

PR #293, opened 2026-09-15
fix(p2p): send range header during URL probes

WindSongShaoGe:
## What

Send `Range: bytes=0-0` during HTTP URL resolution so direct UUID-facade layer opens can complete. Make the UUID facade mock reject GET requests without Range, matching the real facade.

## Why

`open_ro_p2p_uuid` passes `/p2p-uuid/<uuid>` to `open_source_blob_with_size`. For HTTP URLs, `RegistryFsV2` calls `resolve_actual_url_once` before the normal range read. That initial GET previously had no Range header and the UUID facade returned HTTP 400.

This is the direct UUID-facade path. The regular source-registry URL probe does not go through the P2P facade; this PR does not claim otherwise.

## Related issue

Addresses the direct UUID-facade compatibility problem discussed in this PR. No issue-closing claim.

## Scope and non-goals

Only two files change: the URL probe and its UUID facade test mock. The broader OSS snapshot publication, digest lookup, hashing, file-retention, TCP and CI changes have been removed from this PR and preserved separately. The P2P full-chain harness remains on its separate test branch.

- Add a one-byte Range header to the URL probe.
- Return HTTP 400 from the test facade when a GET omits Range, so the existing local-file-missing and remote UUID layer tests exercise the compatibility requirement.

- Public API or generated protocol: no API/schema changes; HTTP probes now carry Range.
- Upgrade and rollback: no migration required.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [ ] `make test-unit`
- [ ] Documentation updated

Commands and results on `51d211c`, based on main `f58d55e`:

```text
cargo test -p overlaybd --lib image::image_file::tests       28 passed
cargo test -p overlaybd --lib backend::registryfs_v2::tests  18 passed
make fmt                                                  PASS
make clippy                                               PASS
```

Negative control: temporarily remove only the production Range fix while keeping the stricter mock. Both `test_missing_file_lower_with_uuid_opens_from_p2p_uuid_facade` and `test_remote_lower_with_uuid_prefers_p2p_uuid_facade` fail; the first reports HTTP 400 while resolving `/p2p-uuid/...`. Restoring the fix makes all 28 image tests pass.

Skipped checks and reasons: full VM/OSS/P2P-chain suites were not rerun for this narrow change. The prior full-chain results apply to the broader follow-up branch, not this two-file fix. No Go, schema, deployment or persisted-format changes.

## Risks and reviewer notes

The Range header applies to HTTP URL probes generally; the registry suite covers existing redirect/authentication behavior. The fix is about URL resolution before range reads, not about missing Range headers in `send_range_request`.

## Checklist

- [x] New behavior is covered by regression tests with a failing negative control.
- [x] Logs and examples contain no production credentials, tokens, or private registry information.
- [x] No generated code was manually edited.


huajq:
This probe talks directly to the registry and never touches the P2P facade — the only requests the facade ever receives already carry an unconditional `Range` header from `send_range_request` — so this change cannot fix the facade rejection it claims to fix; please correct the PR description or provide a repro proving the probe reaches the facade.

WindSongShaoGe:
@huajq Thanks for pointing out the distinction. You're right about the regular registry acceleration path.

The affected case is the direct UUID facade path: `open_ro_p2p_uuid` passes `/p2p-uuid/<uuid>` to `open_source_blob_with_size`, which uses `RegistryFsV2` for HTTP URLs. URL resolution then calls `resolve_actual_url_once` on that facade URL directly, before `send_range_request`.

Without Range, this probe receives HTTP 400 from the UUID facade.

I've added regression coverage in f8631b8 by making the existing mock reject requests without Range. Previously, the mock returned the full blob and masked the issue. With the probe fix temporarily removed, both UUID facade read tests fail; restoring it makes both pass.

All 28 image-layer tests and 18 registry backend tests pass, as do `make fmt` and `make clippy`.

I've also clarified the description to distinguish this UUID path from regular registry acceleration and removed the blanket claim about OSS fallback.

WindSongShaoGe:
I've rebased this PR onto main (f58d55e), retaining the direct UUID-facade Range fix and its regression coverage.
This update also addresses related OSS snapshot P2P issues found during full-chain validation: missing digest-keyed publication, lookup of layers without UUIDs, and retention of transformed upload files through P2P publication.
The full-chain suite is on a separate dependent test branch. Both compression modes passed template launch, snapshot recovery, and OCI conversion reuse, with zero origin data GETs and no repeated OCI conversion. I've updated the PR description to explain the expanded scope and validation.

huajq:
Thanks for digging into the real path — and my mistake: I only considered the registry-URL path and missed the direct UUID-facade path.

huajq (approved):

PR #304, opened 2026-09-19
fix(p2p): publish committed OSS layer artifacts

WindSongShaoGe:
## What

Publish OSS snapshot layers under their committed digest even when local image configs have no descriptor or upload preparation produces different bytes.

Suggested title: `fix(p2p): publish committed OSS layer artifacts`

## Why

Post-commit publication currently scans local image configs. Descriptorless memory layers have no digest to advertise, and compressed or dense-exported upload files are dropped before publication. Consumers cannot discover the committed layer by its digest.

## Related issue

Follow-up to the P2P snapshot publication investigation. PR #293 addressed URL probe Range compatibility; this PR addresses the separate publication gaps.

## Scope and non-goals

Includes committed layer publication and its regression tests. Digest-only OSS reads and the dedicated full-chain E2E harness remain separate changes.

- Return prepared uploads alongside the committed record, keeping transformed temporary files alive through post-commit P2P copies.
- Reuse upload descriptors without another SHA256 pass. Preserve UUID publication and deduplicate digest keys.
- Retain artifacts from the shared ACR export path when OSS uses source-registry publication.
- Keep publication best-effort and strictly after repository success. Preserve startup-recording forwarding and the asynchronous startup-manifest flow.

- Public API or generated protocol: no wire changes; existing repository publish signature remains available, with an additive default trait method for retained uploads.
- Snapshot manifest, artifact layout, or storage format: unchanged; temporary paths are not persisted.
- Upgrade and rollback: existing snapshots remain compatible; rollback removes the additional digest publication.

## Validation

- [x] `make fmt`
- [x] `make clippy`
- [ ] `make test-unit`
- [x] Documentation updated

Commands and results:

```text
cargo test -p agentenv --lib snapshot::
149 passed

cargo test -p agentenv-e2e-tests --test snapshot_oss_e2e_test -- --ignored --test-threads=1
6 passed

make fmt
passed

make clippy

git diff --check
passed
```

On unmodified main 531a002 with only the integration-test changes applied, all three `snapshot_oss_publish*` tests fail because the committed digest is absent from the peer. With the implementation, all pass. Real MinIO/iroh tests compare peer bytes against OSS objects, including compression and sparse dense-export, and verify access after capture cleanup.

Skipped checks and reasons: the broader capability/VM suites and E2B compatibility suite were not run for this publication-only change. No Go or generated schema changes. No performance benchmark was run.

## Risks and reviewer notes

Prepared temporary files now remain until post-commit P2P publication finishes, extending their disk lifetime. Review the ownership transfer through `SnapshotPublication` and the dense-export temporary guard. The read-side digest route is intentionally a separate PR, so this change alone does not establish end-to-end zero-origin restore.

## Checklist


WindSongShaoGe:
CI results checked against base main 531a002:
- unit-tests, fmt, clippy, and agentenv-tests passed.
- Both single-node and Docker Compose E2E passed 16/17 suites. Only 09_e2b_compat failed: Python SDK 2.51.0 and the TypeScript SDK return HTTP 405 during sandbox creation.
- The same failures occur on base main in both environments.
Logs: [PR single-node](https://github.com/kvcache-ai/AgentENV/actions/runs/35425770255/job/105851320437) / [main single-node](https://github.com/kvcache-ai/AgentENV/actions/runs/35417073905/job/105827668514); [PR Docker Compose](https://github.com/kvcache-ai/AgentENV/actions/runs/35425770255/job/105851320419) / [main Docker Compose](https://github.com/kvcache-ai/AgentENV/actions/runs/35417073905/job/105827668617).
These results are consistent with the existing SDK compatibility failure on main.

